PR #7, opened 2021-12-13
Setup Cypress

petterwildhagen:


ReedOnly (changes requested):
Maybe add some small info to readme as well?

petterwildhagen:
> Maybe add some small info to readme as well?

Added info to the readme. Will add more as we move forwards, I am not sure if we will use only Cypress or unit tests with an in memory DB for regression testing, it depends a bit where the functionalty to be tested is

ReedOnly (approved):

PR #13, opened 2021-12-15
Initial header and side menu

denektenina:
The structure is hard coded, but is made to be easy to change as demand changes.

image

### DIFF
--- a/frontend/src/Components/SideMenu/SideMenu.tsx
+++ b/frontend/src/Components/SideMenu/SideMenu.tsx
@@ -1,0 +1,38 @@
+import React from 'react'
+import styled from 'styled-components'
+import ExpandableProject from './ExpandableProject'
+
+const SidebarDiv = styled.div`
+    width: 15rem;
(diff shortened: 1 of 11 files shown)


frontend/src/Components/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+    flex-grow: 1;
+    border-right: 1px solid lightgrey;
+`
+
+export type MenuItem = {
+    name: string
+    cases: string[]
+}
+
+const menuItems = [
```
petterwildhagen, line 17: These menuItems should not be hardcoded, but read from somewhere external, i.e. a database

petterwildhagen: Ok, I am fine with that. Working on it.

ReedOnly (approved):

petterwildhagen (approved):

PR #21, opened 2021-12-15
Setup in memory DB in DCD

petterwildhagen:


denektenina (approved):
The parts I know about look good for a first version

backend/api/api.csproj
```diff
@@ -2,14 +2,17 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
+    <Nullable>disable</Nullable>
```
ReedOnly, line 5: why disable nullable?

backend/api/Services/ProjectService.cs
```diff
@@ -0,0 +1,33 @@
...
+using api.Context;
+using api.Models;
+
+namespace api.Services
+{
+    public class ProjectService
+    {
+        private readonly DCDDbContext _context;
+
+        public ProjectService(DCDDbContext context)
```
ReedOnly, line 13: Think it should be DcdDbContext, to not get too much capital letters

backend/api/Services/Exceptions.cs
```diff
@@ -0,0 +1,18 @@
...
+    public NotFoundInDBException(string message)
+        : base(message)
+    {
+    }
+
+    public NotFoundInDBException(string message, Exception inner)
+        : base(message, inner)
+    {
+    }
+}
```
ReedOnly, line 18: not sure if we need our own DB exception class for now

backend/api/Program.cs
```diff
@@ -22,8 +50,8 @@
     app.UseSwaggerUI();
 }
 
-app.UseAuthentication();
-app.UseAuthorization();
+//app.UseAuthentication();
+//app.UseAuthorization();
```
ReedOnly, line 54: don't uncomment

backend/api/Program.cs
```diff
@@ -2,12 +2,40 @@
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.Identity.Web;
 
+using Microsoft.Data.Sqlite;
+using api.Context;
+using Microsoft.EntityFrameworkCore;
+using api.Services;
```
ReedOnly, line 8: we need proper sorting of dependencies

backend/api/Controllers/ProjectController.cs
```diff
@@ -0,0 +1,37 @@
+using api.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Identity.Web.Resource;
+using api.Models;
+
+namespace api.Controllers;
+
+//[Authorize]
```
ReedOnly, line 9: don't uncomment

backend/api/Services/Exceptions.cs
petterwildhagen: Deleted. Using Exception instead

backend/api/Controllers/ProjectController.cs
petterwildhagen: Fixed

backend/api/Program.cs
petterwildhagen: Fixed

backend/api/Program.cs
petterwildhagen: Fixed

backend/api/api.csproj
petterwildhagen: Jeg tok det bort --- da er det tilsvarende BMT. Fikk warnings som Non-nullable property 'ProjectName' must contain a non-null value when exiting constructor. Consider declaring the property as nullable i Models.cs når jeg hadde det og 
Converting null literal or possible null value to non-nullable type. [api]csharp(CS8600) i projectService.cs.
Ser ut som om default er at Nullable = disabled hvis linjen tas ut helt (slik det er i BMT)

backend/api/Services/ProjectService.cs
petterwildhagen: Fixed

PR #25, opened 2021-12-17
Added charts to project view

denektenina:


frontend/src/types.ts
```diff
@@ -0,0 +1,13 @@
+export type Project = {
```
petterwildhagen, line 1: Will these types be generated as was the case in BMT

denektenina: Yes, eventually. Right now it's the same as in earlier PR's, but moved to a separate file as it is used by several components.

petterwildhagen (approved):
Looks good!!! Like the way testdata has been put into separate files. I guess that models.ts will be generated just as in BMT

PR #39, opened 2021-12-22
Extend data model from spread sheet

petterwildhagen:
Extended data model so that each Project can have mulitple Cases that each contain one Drainage Strategy that contains Production Profile Oil and Production Profile Gas.
Fixed paths in API Urls in Project Controller
Sample Swagger output of Drainage Strategy with Production Profile Oil:
image


backend/api/Models/DrainageStrategy.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public Guid CaseId { get; set; }
+        [ForeignKey("CaseId")]
+        public virtual Case Case { get; set; } = null!;
+        public double NGLYield { get; set; }
+        public virtual ProductionProfileOil ProductionProfileOil { get; set; } = null!;
+        public virtual ProductionProfileGas ProductionProfileGas { get; set; } = null!;
+    }
+
+    public class ProductionProfileOil
```
maiaht, line 19: Why are the classes `ProductionProfileOil` and `ProductionProfileGas` placed in this file? Why are they not placed in separate files?

backend/api/Services/CaseService.cs
```diff
@@ -0,0 +1,33 @@
...
+                {
+                    foreach (Case c in projectCases)
+                    {
+                        DrainageStrategyService drainageStrategyService = new DrainageStrategyService(_context);
+                        DrainageStrategy drainageStrategy = drainageStrategyService.GetDrainageStrategy(c)!;
+                    }
+                }
+                return projectCases;
+            }
+            return null;
```
maiaht, line 30: For `Project` an error is thrown if the project does not exist. Should we not do the same here?

backend/api/Services/DrainageStrategyService.cs
```diff
@@ -0,0 +1,36 @@
...
+                        ProductionProfileGas productionProfileGas = _context.ProductionProfileGas.FirstOrDefault(p => p.DrainageStrategy.Id == drainageStrategy.Id)!;
+                    }
+                    if (_context.ProductionProfileOil != null)
+                    {
+                        ProductionProfileOil productionProfileOil = _context.ProductionProfileOil.FirstOrDefault(p => p.DrainageStrategy.Id == drainageStrategy.Id)!;
+                    }
+                }
+                return drainageStrategy;
+            }
+            return null;
```
maiaht, line 33: For `Project` an error is thrown if the project does not exist. Should we not do the same here?

backend/api/Services/ProjectService.cs
```diff
@@ -12,12 +12,28 @@ public ProjectService(DcdDbContext context)
...
+            if (_context.Projects != null)
+            {
+                if (_context.Cases != null)
+                {
+                    CaseService caseService = new CaseService(_context);
+                    IEnumerable<Project> projects = _context.Projects.ToArray();
+                    foreach (Project p in projects)
+                    {
+                        caseService.GetCases(p);
+                    }
```
maiaht, line 26: Why is this done?

backend/api/Services/ProjectService.cs
```diff
@@ -12,12 +12,28 @@ public ProjectService(DcdDbContext context)
...
+                }
+                return _context.Projects.ToArray();
+            }
+            else
+            {
+                return null;
```
maiaht, line 32: Why not throw an error?

backend/api/Services/ProjectService.cs
```diff
@@ -26,6 +42,11 @@ public Project GetProject(string projectId)
                 {
                     throw new NotFoundInDBException($"Project not found: {projectId}");
                 }
+                if (_context.Cases != null)
+                {
+                    CaseService caseService = new CaseService(_context);
+                    caseService.GetCases(project);
+                }
```
maiaht, line 49: Why is this done?

backend/tests/Services/ProjectService.cs
```diff
@@ -21,7 +23,7 @@ public ProjectServiceTest(DatabaseFixture fixture)
         public void GetQueryable()
         {
             ProjectService projectService = new ProjectService(fixture.context);
-            IQueryable<Project> projectQueryable = projectService.GetAll();
+            IEnumerable<Project> projectQueryable = projectService.GetAll();
```
maiaht, line 26: This test should no longer be called `GetQueryable` and the variable should not be called `projectQueryable`

backend/api/Models/DrainageStrategy.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            get
+            {
+                return YearValue<double>.StrToYearValueArray(InternalTimeSeries);
+            }
+            set
+            {
+                InternalTimeSeries = YearValue<double>.YearValueArrayToString(TimeSeries);
+            }
+        }
```
maiaht, line 36: I don't understand why we have to do this. Why not have a field like:
`public ICollection<TimeSeriesElement> TimeSeries` 
where `TimeSeriesElement` is a class equal to the `YearValue` class that is defined in backend/api/Models/ModelUtils.cs

backend/api/Models/DrainageStrategy.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace api.Models
+{
+
+    public class DrainageStrategy
+    {
+        public Guid Id { get; set; }
+
+        public Guid CaseId { get; set; }
+        [ForeignKey("CaseId")]
+        public virtual Case Case { get; set; } = null!;
```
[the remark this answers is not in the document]
petterwildhagen: Deleted  public Guid CaseId { get; set; }.
Changed to [ForeignKey("Case.CaseId")]

backend/api/Models/DrainageStrategy.cs
```diff
@@ -0,0 +1,61 @@
...
+        public double NGLYield { get; set; }
+        public virtual ProductionProfileOil ProductionProfileOil { get; set; } = null!;
+        public virtual ProductionProfileGas ProductionProfileGas { get; set; } = null!;
+    }
+
+    public class ProductionProfileOil
+    {
+        public Guid Id { get; set; }
+        public Guid DrainageStrategyId { get; set; }
+        public virtual DrainageStrategy DrainageStrategy { get; set; } = null!;
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

backend/api/Models/DrainageStrategy.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+    }
+    public class ProductionProfileGas
+    {
+        public Guid Id { get; set; }
+        public int StartYear { get; set; }
+        public Guid DrainageStrategyId { get; set; }
+        public virtual DrainageStrategy DrainageStrategy { get; set; } = null!;
```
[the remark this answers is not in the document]
petterwildhagen: No, fixed.

backend/api/Models/DrainageStrategy.cs
petterwildhagen: I put them there because they only exist in DrainageStrategy - I can put them in separate classes and then we can see how it looks

backend/api/Services/CaseService.cs
petterwildhagen: A project does not have to contain cases which would be the case when it is initially created so leave as is. 
On another note, the return statment is superflous - the return value is not used -- so I can change the method to void.

backend/api/Services/DrainageStrategyService.cs
petterwildhagen: A case does not have to contain a drainage strategy which would be the situation when the case is first created

backend/api/Services/ProjectService.cs
petterwildhagen: In order to retrieve the cases belonging to a project I needed to run code to extract and then somehow magically (through caching??) the cases are returned with the project. If this is not done then the project is returned without cases even if it contains cases. I left it like this for now to be able to test the model creation

backend/api/Services/ProjectService.cs
petterwildhagen: The method extracts all projects from the DB -- if there are no projects it returns null. A DB with no projects is in and off itself not an error.

backend/api/Services/ProjectService.cs
petterwildhagen: Same reason as above:

In order to retrieve the cases belonging to a project I needed to run code to extract and then somehow magically (through caching??) the cases are returned with the project. If this is not done then the project is returned without cases even if it contains cases. I left it like this for now to be able to test the model creation

backend/tests/Services/ProjectService.cs
petterwildhagen: Fixed

backend/api/Models/DrainageStrategy.cs
petterwildhagen: I can give it a try but I think we need to define TimeSeriesElement for different types because the business wants time series for float and int ,see

https://statoilsrm.sharepoint.com/:x:/r/sites/DCD2/_layouts/15/doc2.aspx?sourcedoc=%7BCEA74DD0-9A05-480F-9888-64B6329AAFE8%7D&file=Data%20model%20DCD.xlsx&action=default&mobileredirect=true&DefaultItemOpen=1&ct=1641209930809&wdOrigin=OFFICECOM-WEB.MAIN.REC-OIB&cid=e19f0657-2a62-40a2-bc8a-fc239afcc849

. Can try this in separate commit. 
That is something like ICollection<YearValue<double>> or ICollection<TimeSeriesElement<double>>

backend/api/Models/DrainageStrategy.cs
petterwildhagen: Fixed

backend/api/Services/ProjectService.cs
petterwildhagen: Fixed

backend/api/Services/ProjectService.cs
petterwildhagen: Completely rewrote code --- fixed

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+                NGLYield = 0.2,
+                Case = Cases[0]
+            };
+            var dstrat2 = new DrainageStrategy
+            {
+                NGLYield = 0.3,
+                Case = Cases[1]
+            };
+            List<DrainageStrategy> d = new List<DrainageStrategy>(new DrainageStrategy[] { dstrat1, dstrat2 });
+            return d;
```
maiaht, line 82: Maybe it is better to not use abbreviations for the variable names?

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+
+        private static List<ProductionProfileOil> GetProductionProfileOil()
+        {
+            var p1 = new ProductionProfileOil
+            {
+                YearValues = new List<YearValue<double>>() { new YearValue<double>(2030, 6.2), new YearValue<double>(2031, 5.3), new YearValue<double>(2032, 4.4) },
+                DrainageStrategy = DrainageStrategies[0]
+            };
+            List<ProductionProfileOil> d = new List<ProductionProfileOil>(new ProductionProfileOil[] { p1 });
+            return d;
```
maiaht, line 93: Maybe it is better to not use abbreviations for the variable names?

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+                DrainageStrategy = DrainageStrategies[0]
+            };
+
+            var p2 = new ProductionProfileGas
+            {
+                YearValues = new List<YearValue<double>>() { new YearValue<double>(2040, 3.2), new YearValue<double>(2041, 3.3) },
+                DrainageStrategy = DrainageStrategies[1]
+            };
+            List<ProductionProfileGas> d = new List<ProductionProfileGas>(new ProductionProfileGas[] { p1, p2 });
+            return d;
```
maiaht, line 110: Maybe it is better to not use abbreviations for the variable names?

backend/tests/Services/ProjectService.cs
maiaht: This has not been fixed. The test should be renamed to something that does not contain the word "queryable" since the function that it tests no longer returns a queryable. The variable on line 26, `projectQueryable`, should also be renamed to something that does not contain the word queryable.

backend/api/Models/YearValue.cs
```diff
@@ -0,0 +1,19 @@
+namespace api.Models
+{
+
+    public class YearValue<T>
+    {
+        public int Id { get; set; }
+        public int Year { get; set; }
+        public T Value { get; set; } = default(T)!;
+
+        public TimeSeriesBase<T> TimeSeriesBase = null!;
```
maiaht, line 10: Why is line 10 included in this class?

maiaht:
image

Why is "case": null under drainageStrategy? Does that mean that you cannot do something like `drainageStartegy.case.id` to get the `id` of the `case` which `drainageStrategy` is a part of?

backend/api/Context/InitContent.cs
petterwildhagen: Fixed.

backend/api/Context/InitContent.cs
petterwildhagen: Fixed

backend/api/Context/InitContent.cs
petterwildhagen: Fixed

backend/api/Models/YearValue.cs
petterwildhagen: Deleted

petterwildhagen:
> image
> 
> Why is "case": null under drainageStrategy? Does that mean that you cannot do something like `drainageStartegy.case.id` to get the `id` of the `case` which `drainageStrategy` is a part of?

It is null because circular references are not resolved (this is configured in Program.cs) so it comes out like this.
I can see if I can remove it in the services code.
No, you can write drainageStrategy.case.Id to get the Id of the case the drainage strategy belongs to.

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+                NGLYield = 0.2,
+                Case = Cases[0]
+            };
+            var drainageStrategy2 = new DrainageStrategy
+            {
+                NGLYield = 0.3,
+                Case = Cases[1]
+            };
+            List<DrainageStrategy> d = new List<DrainageStrategy>(new DrainageStrategy[] { drainageStrategy1, drainageStrategy2 });
+            return d;
```
maiaht, line 82: This should be called `var drainageStrategies`, not `List<DrainageStrategy> d`

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+
+        private static List<ProductionProfileOil> GetProductionProfileOil()
+        {
+            var productionProfileOil1 = new ProductionProfileOil
+            {
+                YearValues = new List<YearValue<double>>() { new YearValue<double>(2030, 6.2), new YearValue<double>(2031, 5.3), new YearValue<double>(2032, 4.4) },
+                DrainageStrategy = DrainageStrategies[0]
+            };
+            List<ProductionProfileOil> d = new List<ProductionProfileOil>(new ProductionProfileOil[] { productionProfileOil1 });
+            return d;
```
maiaht, line 93: This should be called `var productionProfiles`, not `List<ProductionProfileOil> d`

backend/api/Context/InitContent.cs
```diff
@@ -1,30 +1,127 @@
...
+                DrainageStrategy = DrainageStrategies[0]
+            };
+
+            var productionProfileGas2 = new ProductionProfileGas
+            {
+                YearValues = new List<YearValue<double>>() { new YearValue<double>(2040, 3.2), new YearValue<double>(2041, 3.3) },
+                DrainageStrategy = DrainageStrategies[1]
+            };
+            List<ProductionProfileGas> d = new List<ProductionProfileGas>(new ProductionProfileGas[] { productionProfileGas1, productionProfileGas2 });
+            return d;
```
maiaht, line 110: This should be called `var productionProfiles`, not `List<ProductionProfileGas> d`

backend/api/Context/InitContent.cs
petterwildhagen: Fixed

backend/api/Context/InitContent.cs
petterwildhagen: Fixed

backend/api/Context/InitContent.cs
petterwildhagen: Fixed

maiaht (approved):

maiaht (dismissed):

ReedOnly (approved):

PR #48, opened 2022-01-03
Create initial test framework for backend

maiaht:


backend/api/Services/ProjectService.cs
```diff
@@ -24,14 +21,14 @@ public Project GetProject(string projectId)
...
+                var project = _context.Projects.FirstOrDefault(p => p.Id.Equals(projectId));
                 if (project == null)
                 {
-                    throw new Exception($"Project not found: {projectId}");
+                    throw new NotFoundInDBException($"Project not found: {projectId}");
                 }
                 return project;
             }
-            throw new Exception($"Project not found: {projectId}");
+            throw new NotFoundInDBException($"Project not found: {projectId}");
```
petterwildhagen, line 31: This was a bug in the original code. I am not sure if an exception should be thrown if there are no projects in the database, or if just null should be returned. Maybe it makes more sense to just return null and change the method signature to 
public Project? GetProject(string projectId) and return null at the end of the method instean?

petterwildhagen (approved):

petterwildhagen (approved):

PR #66, opened 2022-01-10
Create integration with Org Chart

maiaht:


frontend/src/Services/useService.ts
```diff
@@ -1,20 +1,41 @@
...
                     accessToken,
                 })
+            case 'FusionService':
+                return new __FusionService({
+                    ...config.FusionService,
+                    accessToken: fusionAccessToken,
+                })
             default:
                 break;
         }
```
maiaht, line 41: This does not work. For instance will `ProjectService.getProjectByID(params.projectId!)` fail since `ProjectService` is given by `const ProjectService = useService('ProjectService')` and `useService()` returns the type `__FusionService | __ProjectService | undefined`. The error will then be that `getProjectByID` does not exist on the type `__FusionService`.

maiaht:
I don't know why there are changes in `package-lock.json`. The `package-lock.json` that is shown here is what I get when I do `npm install` based on the `package.json` on master.

frontend/src/Services/FusionService.ts
```diff
@@ -0,0 +1,37 @@
...
+import { config } from "./config"
+import { __BaseService } from "./__BaseService"
+
+export class __FusionService extends __BaseService {
+    getFusionProjects() {
+        return this.get('/contexts?$filter=type%20in%20(%27OrgChart%27)')
+    }
+}
+
+export const useFusionService = () => {
```
aronhoyer, line 14: what's the idea behind sticking with a custom hook here? by doing this, we'll send a network call for every time this hook is called.

frontend/src/Services/FusionService.ts
```diff
@@ -0,0 +1,15 @@
+import { config } from "./config"
+import { __BaseService } from "./__BaseService"
+
+export class __FusionService extends __BaseService {
+    getFusionProjects() {
+        return this.get('/contexts?$filter=type%20in%20(%27OrgChart%27)')
+    }
+}
+
+export const FusionService = () => {
```
aronhoyer, line 10: why are the service classes wrapped in functions? do we need to be able to make new instances for each time we want to consume a service's api?

frontend/src/Services/config.ts
```diff
@@ -7,5 +7,8 @@ export type ServiceConfig = {
 export const config = Object.freeze({
     ProjectService: {
         BASE_URL: `${process.env.REACT_APP_API_BASE_URL}/project`,
+    },
+    FusionService: {
+        BASE_URL: `https://pro-s-context-fprd.azurewebsites.net`
```
aronhoyer, line 12: I assume this won't change between the dev, qa and prod environments?

frontend/src/Components/SideMenu/SideMenu.tsx
```diff
@@ -80,13 +80,13 @@ const SideMenu = () => {
     const navigate = useNavigate()
     const params = useParams()
 
-    const ProjectService = useService('ProjectService')
+    const projectService = ProjectService()
 
     useEffect(() => {
-        if (ProjectService && params.projectId) {
+        if (projectService && params.projectId) {
```
aronhoyer, line 86: checking for whether the service exists or not isn't really necessary.

aronhoyer (changes requested):

frontend/src/Services/FusionService.ts
maiaht: Fixed

frontend/src/Services/FusionService.ts
maiaht: Fixed

frontend/src/Components/SideMenu/SideMenu.tsx
maiaht: Fixed

petterwildhagen (dismissed):
I am not sure if the frontend should integrate with Org chart as we are not sure what role the front end will play in the architecture. It might have to be the back end that integrates with Org chart --- we will know this once we know more about the requirements and hence the over all architecture.

aronhoyer (approved):

PR #67, opened 2022-01-10
Added DataTable for excel-like input and a component for pasting data from excel

denektenina:
The DataTable can handle different vertical and horizontal table headers, but only numbers as cell values right now.

image

By clicking import, we can paste data on a specified format for easy feeding into the data table:

image

The data will replace the existing table (i.e. the horizontal table headers and the data can be replaced, but the horizontal table headers are constant as defined in the hosting component / view)

image

The change has to be actively saved after seeing the result, to avoid overwriting data as an accident:

image


frontend/src/Components/DataTable/helpers.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        expect(replaceOldData(oldGridData, [])[0][1]).toEqual({ value: 123 })
+        expect(replaceOldData(oldGridData, [])[1][2]).toEqual({ value: 928 })
+    })
+
+    test('if the old grid data is "empty", it still inserts the new data', () => {
+        expect(replaceOldData(emptyGrid, changes)[0][1]).toEqual({ value: 111 })
+        expect(replaceOldData(emptyGrid, changes)[1][2]).toEqual({ value: 222 })
+    })
+
+    test('if the old grid data is just an empty array, it returns and empty array', () => {
```
maiaht, line 62: and -> an

frontend/src/Components/DataTable/helpers.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        expect(generateNewGrid(newGridValues, rowTitles).newGridData[0][1].value).toEqual(123)
+        expect(generateNewGrid(newGridValues, rowTitles).newGridData[2][2].value).toEqual(192021)
+    })
+
+    test('Overflowed data is ignored (i.e. there cant be more rows than row titles)', () => {
+        expect(newGridValues.length).toBeGreaterThan(rowTitles.length)
+        expect(generateNewGrid(newGridValues, rowTitles).newGridData.length).toEqual(rowTitles.length)
+    })
+
+    test('If the new data is empty, an array of only the vertical columns is returned', () => {
```
maiaht, line 117: I am not sure if I understand the title of this test. What does "vertical columns" mean? Are not all columns vertical?

frontend/src/Components/DataTable/helpers.ts
```diff
@@ -0,0 +1,47 @@
+import { CellValue } from './DataTable'
+
+export const generateEmptyGridWithVerticalColumns = (rowTitles: string[]) => {
```
maiaht, line 3: I don't understand the name of this function either. Can it be called something like `generateEmptyGridWithRowTitles`?

frontend/src/Components/DataTable/style.css
```diff
@@ -0,0 +1,46 @@
...
+.data-grid-container .data-grid tr th  {
+    padding: 0.25rem 1rem;
+    text-align: right;
+    font-weight: bold;
+}
+
+
+.selected {
+
+}
```
maiaht, line 44: What does it mean that this is empty?

frontend/src/Components/ExcelImport/helpers.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const returnedJsonWhenCapitalLetters = [
+        { hello: '123', test: '222324' },
+        { hello: '456', test: '252627' },
+        { hello: '789', test: '282930' },
+    ]
+
+    test('The data is split into rows by new lines and carriage return feeds and into columns by tabs and commas', () => {
+        expect(tsvToJson(tsvText)).toEqual(returnedJson)
+    })
```
maiaht, line 31: I don't understand the title of this test

frontend/src/Components/ExcelImport/helpers.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    test('If the column headers (the values before the first newline) are fewer than the columns of data in some lines, an array of only the lines with correct amount of columns is returned', () => {
```
maiaht, line 33: Is a "column header" the same as a row title?

frontend/src/Views/CaseView.tsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import { useParams } from 'react-router-dom'
 import styled from 'styled-components'
-import { Tabs, Typography } from '@equinor/eds-core-react'
+import { Button, Tabs, Typography } from '@equinor/eds-core-react'
```
maiaht, line 4: Is it necessary to import the `Button` component in this file? It does not look like it is being used

frontend/src/Components/DataTable/helpers.test.ts
denektenina: yes, maybe not the best name. It's the titles at the first column.

image

The idea is: You have column names, which are at the normal place (top of each column) and then you have vertical column names, that go down the first column. Maybe there is another name for this that is normally used? Do you know?

frontend/src/Components/DataTable/helpers.test.ts
denektenina: fixed

frontend/src/Components/DataTable/helpers.ts
denektenina: Yes, maybe I should use rowTitles everywhere instead of vertical column names

frontend/src/Components/DataTable/style.css
denektenina: Forgot to remove. Fixed now.

frontend/src/Components/ExcelImport/helpers.test.ts
denektenina: It describes how the string is split (you can also look at the code) - \n means new line, so the string is split into rows when it sees that sign, and and \t is just space, so that means new column.

frontend/src/Components/ExcelImport/helpers.test.ts
denektenina: No, column header is the title of each column:

image

frontend/src/Views/CaseView.tsx
denektenina: True, fixed

frontend/src/Components/DataTable/helpers.test.ts
denektenina: Edit: changed it to rowTitles everywhere

maiaht (approved):

PR #100, opened 2022-01-13
Create unit test for project service

petterwildhagen:


backend/tests/Services/ProjectService.cs
```diff
@@ -22,10 +22,15 @@ public ProjectServiceTest(DatabaseFixture fixture)
...
-            IEnumerable<Project> projects = projectService.GetAll();
-
-            Assert.True(projects.Count() > 0);
+            var projectsFromService = projectService.GetAll().OrderBy(p => p.ProjectName);
+            var projectsSourceAndTarget = projects.Zip(projectsFromService);
+            Assert.Equal(projects.Count(), projectsFromService.Count());
+            foreach (var projectPair in projectsSourceAndTarget)
+            {
+                compareProject(projectPair.First, projectPair.Second);
+            }
```
maiaht, line 33: Does it make sense to change this test? Isn't it enough to check that the number of projects is as expected, and then check in the `GetProject`-test that the content of the project is as expected?

backend/tests/Services/ProjectService.cs
```diff
@@ -37,9 +42,80 @@ public void GetProject()
...
+            Assert.Equal(x.WaterInjectorCount, y.WaterInjectorCount);
+            Assert.Equal(x.RiserCount, y.RiserCount);
+            Assert.Equal(x.TemplateCount, y.TemplateCount);
+            Assert.Equal(x.FacilitiesAvailability, y.FacilitiesAvailability);
+            Assert.Equal(x.ArtificialLift, y.ArtificialLift);
+            compareCosts(x.CessationCost, y.CessationCost);
+            compareDrainageStrategy(x.DrainageStrategy, y.DrainageStrategy);
         }
 
+        void compareCosts<T>(TimeSeriesCost<T> x, TimeSeriesCost<T> y)
```
maiaht, line 75: These functions that are used for comparison should follow the same naming convention. Either, all of them should be in the plural version, such as `compareCosts` is now, or all of them should be in the singular version, like `compareCase` is now.

backend/tests/Services/ProjectService.cs
```diff
@@ -37,9 +42,80 @@ public void GetProject()
...
+        void compareYearValues<T>(TimeSeriesBase<T> x, TimeSeriesBase<T> y)
+        {
+            if (x == null || y == null)
+            {
+                Assert.Equal(x, null);
+                Assert.Equal(y, null);
+            }
+            else
+            {
+                var z = x.YearValues.OrderBy(v => v.Year).Zip(y.YearValues.OrderBy(v => v.Year));
```
maiaht, line 111: Maybe `z` could have a more descriptive name?

backend/tests/Services/ProjectService.cs
petterwildhagen: The test framework becomes more data driven - i.e. we check the completeness of all projects that can have different structures - for example some projects can have null for certain elements where this is allowed.

backend/tests/Services/ProjectService.cs
petterwildhagen: Well spotted! Fixed, all functions are now in plural version

backend/tests/Services/ProjectService.cs
petterwildhagen: yearValuePairsXY ?

maiaht (approved):

PR #108, opened 2022-01-17
Introduce API client for Project

aronhoyer:
This pull request introduces:
- A `BaseService` that wraps [Axios](https://axios-http.com)
- A `ProjectService` that implements methods for interacting with the back-end project service
- A `ServicesContext` that exposes an `accessToken` received from MSAL
- A `useService` that consumes the `ServicesContext`, automatically configures a service by a given name and returns that service to the consumer.


frontend/src/Services/config.ts
```diff
@@ -0,0 +1,11 @@
+export type ServiceConfig = {
+    BASE_URL: string
+    accessToken?: string
+    headers?: Record<string, string>
+}
+
+export const config = Object.freeze({
+    ProjectService: {
+        BASE_URL: 'http://localhost:5000/project',
```
ReedOnly, line 9: this needs to be a variable, since it will be different depending on where it it hosted

frontend/src/Components/CasesTable/CasesTable.tsx
```diff
@@ -32,20 +32,20 @@ interface Props {
...
             }
+            // case 'capex': {
+            //     return sort(a.capex, b.capex, sortDirection)
+            // }
+            // case 'drillex': {
+            //     return sort(a.drillex, b.drillex, sortDirection)
+            // }
+            // case 'ur': {
+            //     return sort(a.ur, b.ur, sortDirection)
+            // }
```
ReedOnly, line 46: should this be deleted?

ReedOnly (changes requested):

ReedOnly (approved):

PR #111, opened 2022-01-18
Move drainage strategy under project as asset

petterwildhagen:
Create drainage strategy as asset under project.
Introduce link attribute from Case to drainage strategy.
Created service and controller for drainage strategy. 
Refactor test data generation so that assets are created first, then cases are created.


ericnorway (approved):
Looks good.

PR #118, opened 2022-01-20
added offshore facilities to model. closes issue #89

ericnorway:
Added offshore facilities to model


petterwildhagen (approved):
Looks good. Nicely organized code.

PR #125, opened 2022-01-21
Create model+service+controller for WellProject, update Projct  servi…

petterwildhagen:
Created model for WellProject
Updated sample data for WellProject
Created service and controller for WellProject
Updated ProjectService to return assets along with cases
Updated unit test for Project Service


ericnorway (approved):
Looks good

PR #137, opened 2022-01-25
Delete services and endpoints that have no usecases

petterwildhagen:
Delete services and endpoints that have no usecases


ericnorway (approved):
approved

PR #172, opened 2022-01-27
Create exploration endpoint json approach proof of concept

valentin-krasontovitsch:
ok i think i finally got it. it was a complete pain, and i wasted a lot of time on trying to trick the controller into just getting a string from the body.

i wanted to use this [advice about json and references](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-preserve-references?pivots=dotnet-6-0)
and for that i needed to manually deserialize, mainly as we need to pass an options thingy to the deserializer, and i don't know how to pass that options thingy to the magical hidden controller deserializers.
i finally found a [reference for making controllers accept plain text](https://peterdaugaardrasmussen.com/2020/02/29/asp-net-core-how-to-make-a-controller-endpoint-that-accepts-text-plain/) and applied it. 

now i was able to deserialize manually. and i found this crucial [piece of docs that describes a bit more systematically how this whole reference stuff works](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.referencehandler.preserve?view=net-6.0)
and here we finally have the piece of info which tells us that this approach won't work: it explains when the data is considered well-formed. consider a costprofile (part of an exploration) referencing the exploration and having some yearvalue data. for the exploration in the cost profile, we would want to just use $ref - but according to the docs
A JSON object that contains a $ref metadata property must not contain any other properties.
so it seems to me that this implies that it won't work like that, since the cost profile needs to both refer to its exploration and convey some other data. (kanskje jeg har misforstått - kurset finner sted lørdag og søndag og, ikke sannt?)

i think this shows that the json approach cannot work - we have a create exploration endpoint which manages to deserialize an exploration json that looks like such:

```json
{
  "$id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "project": {
    "id": "0b612717-bebf-4ca7-a4d3-4bdcda47cb61",
    "projectName": "p1",
    "createDate": "2022-01-27T14:40:48.717Z",
    "projectPhase": 0,
    "projectCategory": 0
  },
  "name": "testExploration",
  "wellType": 2,
  "costProfile": {
    "$ref": "1",
    "yearValues": [
      {
        "id": 0,
        "year": 2012,
        "value": 10
      }
    ],
    "epaVersion": "43",
    "currency": 1
  },
  "rigMobDemob": 0
}
```

this one creates an exploration (feel free to inspect it with that fancy VSCode debugger ; )  ) before passing it on to the service, which fails on validation.

in particular, what caught my eye, is that the cost profile seemed to get deserialized to null : /

i think we'll have to use DTOs.

### DIFF
--- a/backend/api/Controllers/ExplorationController.cs
+++ b/backend/api/Controllers/ExplorationController.cs
@@ -1,7 +1,9 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
 using api.Models;
 using api.Services;
 
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Identity.Web.Resource;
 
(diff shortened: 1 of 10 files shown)


valentin-krasontovitsch:
after careful consideration we've decided to go with DTOs

PR #173, opened 2022-01-28
Refactor data model

petterwildhagen:
Moved Articifial Lift from case to drainage strategy, well , surf, topside
Deleted cessation cost on case.
Deleted Measurement class - all values representing physical unit are in base units (m, kg etc)
Moved ProducerCount, GasInjectorCount , WaterInjectorCount from case to Well
Moved FacilitiesAvailability from case to Topside
Moved   RiserCount TemplateCount from case to Surf
Delete the WellType (for the MVP).
Delete units on mass and volume time series
Update test data in sample data generators
Update unit test for project service


ericnorway (approved):
Looks much nicer now with the simplified measurements.

PR #186, opened 2022-02-01
Create client based on swagger

### LINKED ISSUE #68: DCD Frontend: Setup Projects API client

### LINKED ISSUE #114: Frontend: Autogenerate models.ts based on Swagger json

valentin-krasontovitsch:
so i managed to use a tool to auto-generate client code for the frontend.

sadly, my total lack of experience with both react and typescript makes it a wee bit difficult for me to try and inject the generated code into the existing code base.

I would appreciate a bit of help from @aronhoyer or @maiaht, whenever you have time, or you can of course also just look at this on your own first. the generated code lives in `frontend/src/ApiClient`.

closes #114


valentin-krasontovitsch:
after a bit of discussion we have come to the conclusion that we are mostly interested in the model types, as the issue (clearly, duh) states, and not so much in the API client code. we should investigate whether we can only create the model types, or simply remove the client code.

the argument against the generated client code is its monolithic nature - one big api client for all resources. we have chosen a more granular approach. perhaps the auto-generator can be tweaked into mirroring our approach?

ericnorway:
Decided to use dtsgenerator package instead

PR #203, opened 2022-02-02
Add create topside endpoint

valentin-krasontovitsch:
Please take a look at this, I've followed mostly the patterns from the Surf, with a bit of a twist here and there, @petterwildhagen will be able to illuminate. Feel free to take over this code, since i won't be able to react / comment before i return from my vacation.


ericnorway:
added with PR 218

PR #208, opened 2022-02-03
Add controller and service for case

petterwildhagen:


backend/api/Services/ProjectService.cs
```diff
@@ -106,5 +109,20 @@ public void AddSurfsToProject(Project project, Surf surf)
         {
             project.Surfs.Add(surf);
         }
+
```
ericnorway, line 112: these aren't needed anymore

petterwildhagen: removed

ericnorway (approved):
looks good

PR #212, opened 2022-02-03
Adding service and controller for Transport 

eiriksanderfjeld:
Service, Controller, Dto, and Controller updated with the correct values #157

### DIFF
--- a/backend/api/Controllers/TransportsController.cs
+++ b/backend/api/Controllers/TransportsController.cs
@@ -1,3 +1,8 @@
+using api.Adapters;
+using api.Dtos;
+using api.Models;
+using api.Services;
+
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Identity.Web.Resource;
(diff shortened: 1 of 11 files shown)


backend/api/Services/TransportService.cs
```diff
@@ -5,14 +5,47 @@
...
+                throw new Exception(); //TODO FIX EXCEPTION
+            }
+        }
+
+
+        public bool DeleteTransport(Transport transport)
+        {
+            if (_context.Transports != null)
+            {
+                _context.Transports.Remove(transport);
```
ericnorway, line 39: when updating or deleting data in EF, we first need to get the entity from the context. Otherwise EF will think it's a different object with the same key and throw unique key constraint errors.

            var transport = GetTransport(transportId);
            _context.Transports.Remove(transport);
            _context.SaveChanges();

https://docs.microsoft.com/en-us/ef/core/saving/basic

backend/tests/Services/SurfServiceTest.cs
```diff
@@ -31,8 +31,8 @@ public void GetAllSurf()
             ProjectService projectService = new ProjectService(fixture.context);
             SurfService surfService = new SurfService(fixture.context, projectService);
             var project = projectsFromSampleDataGenerator.First();
-            var surfsFromProject = surfService.GetSurfs(project.Id);
-            // var surfsExpectedActual = projectsFromSampleDataGenerator.Zip(surfsFromProject);
+            surfService.GetSurfs(project.Id);
+
```
ericnorway, line 35: I have tests for the different CRUD operations for drainage strategy and substructure if you want to add those now.

backend/api/Controllers/TransportsController.cs
```diff
@@ -11,10 +16,45 @@ namespace api.Controllers
...
+        public Transport UpdateTransport([FromBody] TransportDto transportDto)
+        {
+            var transport = _transportAdapter.Convert(transportDto);
+            return _transportService.UpdateTransport(transport);
+        }
+
+        [HttpPost("{projectId}", Name = "transport")]
+        public Transport CreateTransport([FromBody] TransportDto transportDto)
+        {
+            Console.WriteLine(transportDto);
```
ericnorway, line 41: remove console.writeline

backend/api/Services/ProjectService.cs
```diff
@@ -106,5 +106,10 @@ public void AddSurfsToProject(Project project, Surf surf)
         {
             project.Surfs.Add(surf);
         }
+
```
ericnorway, line 109: Not necessary to add new methods to project service now. in the transportservice, just set the transport.Project = project, and EF will automatically add the transport to the project. See SubstructureService.CreateSubstructure().

backend/api/Services/TransportService.cs
```diff
@@ -5,14 +5,47 @@
...
         private readonly DcdDbContext _context;
+        private readonly ProjectService _projectService;
 
-        public TransportService(DcdDbContext context)
+        public TransportService(DcdDbContext context, ProjectService projectService)
         {
             _context = context;
+            _projectService = projectService;
         }
+        public Transport CreateTransport(Transport transport)
```
ericnorway, line 18: Return the whole project for create, update, and delete operations.

backend/api/Services/TransportService.cs
eiriksanderfjeld: Done!

backend/api/Controllers/TransportsController.cs
eiriksanderfjeld: doh

backend/tests/Services/SurfServiceTest.cs
eiriksanderfjeld: Great, added them as template

PR #222, opened 2022-02-04
Fix drainage strategy test to include case id

ericnorway:
Fix drainage strategy test to include case id


petterwildhagen (approved):
looks good!

PR #244, opened 2022-02-08
Cleanup services

petterwildhagen:


eiriksanderfjeld (dismissed):

backend/api/Services/TransportService.cs
```diff
@@ -15,50 +15,59 @@ public TransportService(DcdDbContext context, ProjectService projectService)
...
+        {
+            foreach (Case c in _context.Cases!)
             {
-                return new Transport();
+                if (c.TransportLink == transportId)
+                {
+                    c.TransportLink = Guid.Empty;
+                }
             }
+            _context.SaveChanges();
```
ericnorway, line 70: Don't need to save the context again. It is also being saved in DeleteTransport().

eiriksanderfjeld (approved):

PR #247, opened 2022-02-08
POC for API request/response tests

valentin-krasontovitsch:
i managed - just had to expose the `Program` class in `Program.cs`, change the SDK and update a dependency version used in the testing project, and then i could write a test class that spins up a test server (in memory!) which can be queried.

i very much followed the [official docs for integration tests](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-6.0#mock-authentication), and i think i've looked at them a few weeks ago, but i just wasn't ready yet then...

anywho we have a testserver, with a dummy authenticatior injected, and a simple request against projects, which returns good stuff (based on the logs during the test).

Haven't checked yet what made us pursue this path, that's gonna come tomorrow hopefully.

ping @ericnorway @petterwildhagen 

I hope this is usable (to detect the weird behaviour with attach / include), and it's definitely extendable to properly check all the things)

### DIFF
--- a/backend/api/Program.cs
+++ b/backend/api/Program.cs
@@ -110,3 +110,5 @@ app.UseCors(_accessControlPolicyName);
 app.MapControllers();
 
 app.Run();
+
+public partial class Program { }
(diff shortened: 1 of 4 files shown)


petterwildhagen (approved):
Looks good!!!

PR #250, opened 2022-02-09
autogenerated

ericnorway:


frontend/src/Views/ProjectView.tsx
```diff
@@ -46,13 +46,24 @@ const ProjectView = () => {
     if (!project) return null
 
     const dataX: string[] = []
-    let dataProdProfileGas: number[] = []
-    let dataProdProfileOil: number[] = []
+    let dataProdProfileGas: number [] = []
```
aronhoyer, line 49: shouldn't have spacing between `number` and `[]`

aronhoyer, line 49: there shouldn't be a space between `number` and `[]`

frontend/src/Utils/common.tsx
```diff
@@ -0,0 +1,5 @@
+import Project = Components.Schemas.Project;
+
+export function GetDrainageStrategy(project: Project, drainageStrategyId: string | undefined) {
```
aronhoyer, line 3: `arg: string | undefined` can be simplified to `arg?: string`

aronhoyer (changes requested):
looks good overall. added a couple of comments on code style

frontend/src/Views/ProjectView.tsx
ericnorway: fixed

frontend/src/Utils/common.tsx
ericnorway: fixed

frontend/package.json
```diff
@@ -37,7 +38,8 @@
         "test": "react-scripts test",
         "cyopen": "./node_modules/.bin/cypress open",
         "cyrun": "./node_modules/.bin/cypress run",
-        "test:ci": "npm run test -- --watchAll=false --coverage src --coverageDirectory=coverage"
+        "test:ci": "npm run test -- --watchAll=false --coverage src --coverageDirectory=coverage",
+        "autogenerate-models": "npx dtsgen --out ./src/types.d.ts --url https://ase-dcd-backend-dev.azurewebsites.net/swagger/v1/swagger.json"
```
aronhoyer, line 42: no need for `npx` in this script. you can simply call `dtsgen` directly in the script :+1:

frontend/src/Views/ProjectView.tsx
```diff
@@ -49,10 +49,21 @@ const ProjectView = () => {
...
-        dataProdProfileOil = c.drainageStrategy?.productionProfileOil?.yearValues?.map(v => v.value)
-        dataX.push(c.name)
+    project.cases?.forEach(c => {
+        const drainageStrategy = GetDrainageStrategy(project, c.drainageStrategyLink);
+        if (drainageStrategy === undefined)
+          return;
+        if (drainageStrategy.productionProfileGas !== undefined 
+            && drainageStrategy.productionProfileGas !== null
+            && drainageStrategy.productionProfileGas.yearValues !== undefined
+            && drainageStrategy.productionProfileGas.yearValues !== null)
```
aronhoyer, line 59: no need to fix this right now, but this can be severely simplified to 
```js
if (drainageStrategy.productionProfileGas?.yearValues) {
  /* code */
}
```
this is because `null` and `undefined` are falsy values. no need to explicitly check for `undefined` or `null`. additionally of note, is the `?.` object access syntax. this is called [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). I believe C# has this as well.

aronhoyer (changes requested):

frontend/package.json
ericnorway: removed npx

PR #291, opened 2022-02-18
wip add new project

ericnorway:


frontend/src/Components/Modal.tsx
```diff
@@ -0,0 +1,47 @@
...
+        closeModal(false);
+    }
+
+    if (!isOpen) return null;
+    return (
+        <Portal>
+            <ModalDiv>
+                <div>{children}</div>
+                <div>
+                    <Button onClick={handleOkClick}>{okButtonText}</Button>
```
aronhoyer, line 39: i'm not sure we necessarily want only the `onClick` handler and button texts. i think we're better served with passing the actual button as props (i.e. we pass a component as a prop)

frontend/src/Utils/common.tsx
```diff
@@ -1,3 +1,119 @@
-export function GetDrainageStrategy(project: Components.Schemas.ProjectDto, drainageStrategyId?: string) {
-  return project.drainageStrategies?.find(o => o.id === drainageStrategyId);
+export const GetDrainageStrategy = (project: Components.Schemas.ProjectDto, drainageStrategyId?: string) => {
+    return project.drainageStrategies?.find(o => o.id === drainageStrategyId);
+};
+
+const enum ProjectPhase {
+    Null = 0,
```
aronhoyer, line 6: you don't need to explicitly assign values to enum keys if the values are just `0..n`. that's the default 😄

frontend/src/Views/CreateProjectView.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const updateNameHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setName(event.target.value);
+    }
+
+    const updateDescriptionHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setDescription(event.target.value);
+    }
+
+    useEffect(() => {
+        if (commonLibraryService) {
```
aronhoyer, line 78: no need for this check

frontend/src/Services/ProjectService.ts
```diff
@@ -9,6 +9,10 @@ export class __ProjectService extends __BaseService {
     getProjectByID(id: string) {
         return this.get(`/${id}`)
     }
+
+    postProject(project: Components.Schemas.ProjectDto) {
```
aronhoyer, line 13: let's try to stier away from using http verbs in service methods. "insert" or "create" are better options.

frontend/src/Services/__BaseService.ts
```diff
@@ -26,21 +26,28 @@ export class __BaseService {
         this.client.defaults.headers.common = {
             Accept: 'application/json',
             Authorization: `Bearer ${config.accessToken}`,
+            'Content-type': 'application/json',
```
aronhoyer, line 29: there's no need to explicitly set this. axios should be able to infer this based on the request body type. that being said, body (as defined on line 34) should not be a string, but passed raw. there's also a change to this coming with my branch on the ProjectView implementation, so you don't necessarily need to update this right now. have a look at [the commit](https://github.com/aronhoyer/dcd/commit/1b5b916c102e2f5fefcbc2e533059d3459ecb285) if you want to sync with how i've implemented the changes to the BaseService

frontend/src/Views/CreateProjectView.tsx
```diff
@@ -0,0 +1,154 @@
...
+const grey = tokens.colors.ui.background__scrim.rgba;
+
+interface Props {
+	isOpen: boolean;
+	closeModal: Function;
+}
+
+const CreateProjectView = ({ isOpen, closeModal }: Props) => {
+    const [projects, setProjects] = useState<Components.Schemas.CommonLibraryProjectDto[]>();
+    const [selectedProject, setSelectedProject] = useState<Components.Schemas.CommonLibraryProjectDto>();
```
aronhoyer, line 31: let's look into implementing [react hook form](https://www.npmjs.com/package/react-hook-form) instead of handling form state ourselves. forms re-evaluations and re-renders can quickly get out of hand.

frontend/src/Views/CreateProjectView.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }, []);
+
+    if (!projects) return null;
+
+	return (
+		<Modal isOpen={isOpen} closeModal={closeCreateProjectView} okButtonText='Create Project' cancelButtonText='Cancel'>
+            <div>
+                <Typography variant='h1'>Create Project</Typography>
+                <ProjectSelect>
+                    <Icon data={search} color={grey}></Icon>
```
aronhoyer, line 97: when a component doesn't have children, it's considered self closing, so instead of writing `<Icon></Icon>` we can write `<Icon />`

frontend/src/Views/CreateProjectView.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    <ProjectDropdown
+                        id="textfield-name"
+                        label={'CommonLib project'}
+                        placeholder={'Select a CommonLib project'}
+                        items={projects!.map(p => p.name!)}
+                        handleSelectedItemChange={(changes: UseComboboxStateChange<string>) => onSelected(changes.selectedItem)}
+                    />
+                </ProjectSelect>
+                <div>
+                    <Label htmlFor="textfield-name" label="Name" />
```
aronhoyer, line 107: prefer `<TextField />` over `<Label />` and `<Input />`. take a look at the [eds storybook](https://eds-storybook-react.azurewebsites.net/?path=/story/introduction--page)

frontend/package.json
```diff
@@ -15,6 +15,7 @@
         "react": "^17.0.2",
         "react-datasheet": "^1.4.9",
         "react-dom": "^17.0.2",
+        "react-focus-lock": "^2.8.1",
```
aronhoyer, line 18: prefer [react focus on](https://www.npmjs.com/package/react-focus-on)

aronhoyer (changes requested):

frontend/src/Views/CreateProjectView.tsx
ericnorway: Will look into this after the demo.

frontend/src/Services/__BaseService.ts
```diff
@@ -30,17 +32,22 @@ export class __BaseService {
...
         })
         return data
     }
 
-    protected get(path: string, options?: RequestOptions) {
+    protected get<T = any>(path: string, options?: RequestOptions): Promise<T> {
         return this.request(path, { ...options, method: 'GET' })
     }
+
+    protected post<T = any>(path: string, body?: Record<string, any>, options?: RequestOptions): Promise<T> {
```
aronhoyer, line 50: `body` is still in the parameter list. since it's accessible through the `options` object, the `body` parameter is superfluous. removing it from the parameter list also allows for not having to explicitly pass it on in the `this.request` call on line 51.

aronhoyer (changes requested):
looks good overall. just one more change and we can merge it. 👍

aronhoyer (approved):

aronhoyer (approved):

PR #295, opened 2022-02-20
E2e tests

valentin-krasontovitsch:


valentin-krasontovitsch:
Ready for review!

I'm sorry, there's quite a bit that has happened in this PR - but I grouped the changes into hopefully kindof atomically scoped commits, so if you don't just look at the changes but rather look at the commits and their respective changes, it should be easier to understand what's going on.

also the first couple commits up to the e2e tests we have already discussed : )

i'll be available for questions and comments or discussions

backend/api/Adapters/ProjectAdapter.cs
```diff
@@ -15,6 +15,7 @@ public static Project Convert(ProjectDto projectDto)
             var project = new Project();
             project.Name = projectDto.Name;
             project.CommonLibraryId = projectDto.CommonLibraryId;
+            project.CreateDate = DateTimeOffset.UtcNow;
```
petterwildhagen, line 18: Only set create date if not already set. Otherwise the date is updated everytime this method gets called (creating an object to send to the backend).

petterwildhagen (changes requested):
Excellent!!!
Just fix the minor bug with the create date when returning the project from the project Dto.
There are also some codacy errors that should be fixed.

backend/api/Adapters/ProjectAdapter.cs
valentin-krasontovitsch: very good point, thank you!

actually i'm wondering... which create day exactly are we trying to model here? the date on which a project gets created in our app? or the date on which a project was created according to commonlib? maybe we can discuss this a bit on monday

because if we say we only set the date if it's not set on the project dto, then we should make it optional on the project dto. then the question becomes in which cases do we expect to receive a created date in the project dto? if we only ever receive a project dto as part of the create project endpoint, and if we say that all projects have to come from commonlib do we sometimes import a creation date from commonlib, sometimes not? but if the adapter possibly gets used by other endpoints as well, then we should maybe move the setting of the date for the create project use-case to the service or the controller?!

petterwildhagen: Correction: The create date has to be set -- obviously -- be set when project is created.

backend/tests/Adapters/DrainageStrategyAdapterShould.cs
```diff
@@ -17,11 +17,10 @@ public void ConvertDrainageStrategyDtoToDataModel()
         {
```
petterwildhagen, line 17: This file can be deleted

backend/api/Adapters/ProjectAdapter.cs
petterwildhagen: So simply set the date here:
image

valentin-krasontovitsch: fixed!

backend/tests/Adapters/DrainageStrategyAdapterShould.cs
valentin-krasontovitsch: I disagree with deleting any unit tests for now - we have not covered all our bases with the end to end tests yet. we need both more test scenarios, and more varied test data before we can be sure that all services and adapters all covered by e2e tests, and then we can delete the unit tests which at the moment cover services and adapters.

valentin-krasontovitsch:
@petterwildhagen i btw also checked if these changes have any effects on the frontend by rebuilding and comparing the automatically generated models file - they don't, the frontend should be unaffected by the changes in this PR (i mean it won't break anything)

petterwildhagen (approved):
Looks good!

PR #313, opened 2022-02-22
Run lint on commit

aronhoyer:
This PR introduces:
- Eslint as a replacement of Prettier
- Husky, used to run Lint Staged—which in turn runs Eslint on files staged in git on each commit

Eslint will only exit with a non-0 status code if formatting issues can't be automatically fixed.


ericnorway (commented):
The changes look good, but codacy is complaining about some lines being too long. Also it doesn't like console.log for some reason.

ericnorway (approved):
Looks good.

PR #353, opened 2022-03-02
Simplify launchsettings

valentin-krasontovitsch:
for local development of the backend it turns out that we can populate most settings just from the app configuration resource, assuming that we are authenticated in the command line with the `az` CLI.

removed stuff from the launch settings template and described necessary steps in the readme.


petterwildhagen (approved):
Looks good!!

PR #357, opened 2022-03-02
Update recent projects storage strategy

### LINKED ISSUE #296: DCD frontend: update storage strategy for recently viewed projects
at the moment when a user views a project, we note for displaying recent projects the project id together with a timestamp.

then, to display the recently visited projects, we fetch all such key value pairs for projects fetched from the backend end (so far they exist), sort them, and display the desired information.

an alternative approach during a user's visit of a project is to
- store all information needed to display the recent project card
- organize the information in an array (sorted by FIFO)
- keep the array at a max length of 5

that way the info is sorted, and all info needed to display recent projects is available independent of the backend.

valentin-krasontovitsch:
closes #296


frontend/src/Components/RecentProjects.tsx
```diff
@@ -68,7 +68,7 @@ function RecentProjects({ projects }: Props) {
             ? tokens.colors.ui.background__info.rgba
             : tokens.colors.ui.background__danger.rgba
 
-        const labelText = phase ?? "TBD"
+        const labelText = phase || "TBD"
```
aronhoyer, line 71: for this case this might be what we want, but something to keep in mind with the logical or operator, is that a lot of values are evaluated as `false` in javascript. for example, and empty string is evaluated as `false` (0 is as well), which in some cases might lead to unintended behaviour of our program. for those cases we can use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). just something to keep in mind :)

frontend/src/Utils/common.tsx
```diff
@@ -1,24 +1,35 @@
...
-    return `DG${(phaseNumber + 1).toString()}`
-}
-function recentProjectKey(projectId: string) {
-    return recentProjectKeyPrefix + projectId
-}
-export function StoreRecentProject(projectId: string) {
-    const timeStamp = new Date().getTime()
-    const key = recentProjectKey(projectId)
-    localStorage.setItem(key, timeStamp.toString())
+export function RetrieveRecentProjects() {
```
aronhoyer, line 11: this can be a static method on the new `RecentProject` model

frontend/src/Utils/common.tsx
```diff
@@ -1,24 +1,35 @@
...
+    const recentProjectJSON = localStorage.getItem(recentProjectsKey)
+    const recentProjects: RecentProject[] = JSON.parse(recentProjectJSON ?? "[]")
+    return recentProjects
 }
-export function RetrieveLastVisitForProject(projectId: string) {
-    const timeStamp = localStorage.getItem(recentProjectKey(projectId))
-    return timeStamp
+export function StoreRecentProject(project: Project) {
```
aronhoyer, line 16: this can be an method on the new `RecentProject` model

aronhoyer (changes requested):
i don't know about having a separate model for a recently visited project. essentially it's just a project, so by having two models, we're basically duplicating code and maintenance.

what was your reasoning behind creating a separate model for it?

frontend/src/Components/RecentProjects.tsx
valentin-krasontovitsch: thanks, will keep it in mind!

valentin-krasontovitsch:
> what was your reasoning behind creating a separate model for it?

@aronhoyer one of the goals with the restructuring was to store only the information needed for the recent project card - rather than explicitly listing the properties we want when we store, retrieve, and render, it seemed elegant to gather the information. maybe it shouldn't be its own "model", just a type, or class, or something, but i'm not familiar with the naming conventions...
i can break it up into the particular fields though, but i think it makes the code more cumbersome?
do you have a suggestion for solving this?

ahmedoabdi:
Any thoughts or updates regarding the last comment @aronhoyer?

aronhoyer:
> do you have a suggestion for solving this?

i don't know if i said this, but in order to get around the duplicate model issue, we could store an array of Projects. to reduce the payload size, maybe we could have a generalized util function to clear out object fields that have nullish or falsy values.

valentin-krasontovitsch:
ok cool i can do that, for sure. i'm gonna 

- store projects in the array
- make the store and retrieve functions methods on the project type

but i'm a bit uncertain as to what you exactly mean by clearing out object fields that have falsy or nullish values - i thought maybe just stripping assets, as they are not used in this context and are potentially large, as in settings those to `null`, but that's not what you mean, right?

aronhoyer:
@valentin-krasontovitsch not what i meant, but we should definitely strip any values from the object that are not pertinent to displaying a recently visited project

valentin-krasontovitsch:
OK, gotcha I think.

another issue - it seems like the "complex" types don't properly survive being stored as JSON and then parsed again - in the code, everything is fine typewise, but when I try to for instance run the `toString` function on an instance of `ProjectPhase`, I just get `[object Object]` instead of the expected. Also, the createdAt does not come back as a `Date`, but rather as a string after parsing it from JSON.

any pointers, @aronhoyer?

aronhoyer:
use `JSON.stringify` instead of `toString` :)

valentin-krasontovitsch:
oh i use `stringify` when storing the projects - `toString` is a method of the projectphase type, which translates the number to a phase string. it seems like i had to use custom parsing of JSON to get back `Date` and `Phase`, passing an extra function as second argument to `JSON.parse` when retrieving the stringified projects from localstorage. gonna push the new changes soon

frontend/src/Utils/common.tsx
valentin-krasontovitsch: done on Project model

frontend/src/Utils/common.tsx
valentin-krasontovitsch: done on Project model

aronhoyer (approved):

PR #370, opened 2022-03-08
Case 363 and 365. Changed case view and added "create" and "link" asset buttons.

DanielBohme:
Scrapped old case view with dummy data and added starting point for Aron's Figma concept for case view. Added non editable DG4 field and description field. Also added buttons for "Create an asset" and "Link to asset" with forms (no data).

### DIFF
--- a/frontend/src/Views/CaseView.tsx
+++ b/frontend/src/Views/CaseView.tsx
@@ -1,17 +1,30 @@
-import { Tabs, Typography } from "@equinor/eds-core-react"
-import { useEffect, useState } from "react"
+import {
+    Tabs,
+    Typography,
+    Button,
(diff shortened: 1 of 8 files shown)


petterwildhagen (approved):
Looks good!

frontend/src/Views/CaseView.tsx
```diff
@@ -22,13 +35,40 @@ const CaseViewDiv = styled.div`
 
 const CaseHeader = styled(Typography)`
     margin-bottom: 2rem;
+    display: flex;
+
```
ManveerPandher, line 39: whitespace

ManveerPandher (approved):

PR #372, opened 2022-03-08
Update modal implementation

aronhoyer:
This PR updates the modal implementation.

The plan was initially to use EDS' Dialog component, but the documentation seems to be behind the actual implementation. Additionally, it was missing parts that are pertinent to a functional dialog/modal component. As such, we're moving ahead with our own custom modal implementation, which seems to work better than EDS' Dialog.

The modal is currently implemented such that it is conditionally rendered by the consumer, and is automatically mounted in a React Portal. Additionally, it is implemented with focus lock functionality (meaning the user can't tab out of the modal when it's mounted) for accessibility concerns.

### DIFF
--- a/frontend/src/Components/Header.tsx
+++ b/frontend/src/Components/Header.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from "react"
+import { useState } from "react"
 import styled from "styled-components"
 import { Typography, Button } from "@equinor/eds-core-react"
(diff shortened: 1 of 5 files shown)


eiriksanderfjeld:
@ManveerPandher Vil du se over denne PR'en og se om vi kan ta den inn?

ManveerPandher:
Tar en titt

PR #373, opened 2022-03-09
Create drainage strategy

aronhoyer:
this pr introduces a form to create a drainage strategy, and a component to map asset types to the correct forms

depends on #372


eiriksanderfjeld:
@ManveerPandher Vil du ta en kikk på denne saken også?

PR #377, opened 2022-03-09
Feature/362 add update on case api

ahmedoabdi:
- Added update method to Case API with PUT
- Changed CaseAdapter to non static class
- Added UpdateCase method to CaseService without CopyData method
-  Added test for UpdateCase

### DIFF
--- a/backend/api/Adapters/CaseAdapter.cs
+++ b/backend/api/Adapters/CaseAdapter.cs
@@ -4,20 +4,20 @@ using api.Services;
 
 namespace api.Adapters
 {
-    public static class CaseAdapter
+    public class CaseAdapter
     {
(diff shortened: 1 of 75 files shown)


petterwildhagen (changes requested):
Looks good but keep CaseAdapter static

backend/api/Adapters/CaseAdapter.cs
```diff
@@ -4,20 +4,20 @@
 
 namespace api.Adapters
 {
-    public static class CaseAdapter
+    public class CaseAdapter
```
petterwildhagen, line 7: keep static

petterwildhagen (changes requested):

petterwildhagen (approved):
Looks good!

PR #379, opened 2022-03-09
Added filtering to projects from commonlib

fredrikleknes:
Added filtering for projects from commonlib


eiriksanderfjeld:
Fikser du kommentaren fra codacy, så ser det bra ut.

petterwildhagen (approved):
Looks good!

PR #381, opened 2022-03-09
Feature/remove copydata api

ahmedoabdi:
This PR removes CopyData from the different asset services in the backend as it is not necessary after changes done on the update action in the controllers.

The update action itself has been updated to only take in the asset dto as a parameter.


petterwildhagen (approved):
Looks good!

PR #384, opened 2022-03-11
Added DG1, 2, 3 to model

fredrikleknes:


frontend/src/models/Case.ts
```diff
@@ -2,6 +2,9 @@ export class Case {
     capex?: number
     createdAt?: Date | null
     description?: string
+    firstDecisionGate?: Date | null
```
petterwildhagen, line 5: Is firstDecisionGate a domain concept from the business? Why not use DG1Date ? We (the technical team) should not introduce new domain names (or concepts in general).

fredrikleknes: The reason i put it like that was since DG4 is declared lastDecisionGate, and DG does stand for decisiongate, no? But i do agree zeroethDecisionGate sounds silly. I'll change it

petterwildhagen: Yes. DG = decision gate. But I think DG is a commonly understood term among the users, so we can leave it as is.
See also here http://www.ipt.ntnu.no/~stanko/files/Courses/TPG4230/2020/Lecture_notes/20200113.pdf - slide no 1.
If you just put D1Date for firstDecisionGate it is fine.

frontend/src/models/Case.ts
```diff
@@ -2,6 +2,9 @@ export class Case {
     capex?: number
     createdAt?: Date | null
     description?: string
+    DG1Date?: Date | null
+    DG2Date?: Date | null
+    DG3Date?: Date | null
     lastDecisionGate?: Date | null
```
petterwildhagen, line 8: Endre lastDecisionGate til DG4Date så er alt OK. Også på linje 22 under

fredrikleknes: Aight. Done. Forresten, så du i types filen, så ble det autogenerert en oppdatering av 'updatecase'. Jeg er litt usikker på hvorfor den ble generert, men tok den med

petterwildhagen (approved):
Looks good!

PR #388, opened 2022-03-11
Move call to adapter to service

petterwildhagen:


ahmedoabdi (approved):
Nice work, looks good.

PR #394, opened 2022-03-14
Fixed create case bug

petterwildhagen:


fredrikleknes (approved):
Looks good! Exactly the same solution as me and Tord came up with

PR #405, opened 2022-03-15
Added functional date field for DG4

DanielBohme:
Still some bugs with submitting new date, but adding PR as DG4 date is mapped to case.DG4.

Current bugs:
1. When changing date, first change is undefined, second change is defined but uses input on first change. So date always lags behind on change. Needs to be fixed at some point.
2. When changing date, date isn't updated on field after change. This is due to value retrieving case.DG4 date on load currently. Also needs to be fixed at some point.


petterwildhagen (approved):
Looks good!

PR #409, opened 2022-03-16
Added dropdowns for linking assets

taustad:
Removed buttons for adding and linking assets
Added dropdowns for linking assets


frontend/src/Utils/CreateAsset.ts
```diff
@@ -0,0 +1,92 @@
+import { Project } from "../models/Project"
```
petterwildhagen, line 1: Avoid adding code to add assets / data in utilities folder. 
Do not create default assets with dummy data.

frontend/src/Utils/CreateAsset.ts
```diff
@@ -0,0 +1,92 @@
...
+import { GetDrainageStrategyService } from "../Services/DrainageStrategyService"
+import { GetExplorationService } from "../Services/ExplorationService"
+import { GetSubstructureService } from "../Services/SubstructureService"
+import { GetSurfService } from "../Services/SurfService"
+import { GetTopsideService } from "../Services/TopsideService"
+import { GetTransportService } from "../Services/TransportService"
+import { GetWellProjectService } from "../Services/WellProjectService"
+
+export const CreateAsset = async (assetType: string, caseId: string, projectId: string) => {
```
petterwildhagen, line 10: When clicking Create Asset, the user is navigated to a new window where the asset is created.

petterwildhagen (changes requested):
The implementation of the linking work but discuss with @ManveerPandher who removed 'links' from the case domain model
image

Do a bit less in the PR. Do not create the assets. Delete all code associated with that.

petterwildhagen (approved):
Looks good!

PR #412, opened 2022-03-17
Feature/385 create asset list in case view

ManveerPandher:
What has been done within this PR:
- Duplicated DrainageStrategyView, and created the 6 other assets: "ExplorationView", "SurfView", "SubstructureView", "SurfView", WellProjectView" and "TransportView"
- DTO's has been created for the assets, so the Project.ts datamodel now also correctly parse / map their related asset-data. 

Why this has to be done:
- The views may be used in other Views in order to be populated and display data from their representative asset models


frontend/src/Views/TopsideView.tsx
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const saveDataImport = () => {
+        // TODO CODE TO SAVE DATA HERE
+        setDataIsChanged(false)
+    }
+
+    return (
+        <Wrapper>
+            <ImportButton onClick={openImportView}>Import...</ImportButton>
```
petterwildhagen, line 163: Delete import button. Data for time series is entered by pasting in array and specifying a start year in a separate field

petterwildhagen (changes requested):
Delete all Import ... of tables since the time series will be entered as arrays for each time series individually.

petterwildhagen (approved):
Looks good!

PR #415, opened 2022-03-17
Create excel export to STEA

petterwildhagen:
PR introduces export of export to STEA.
Can be tested in swagger where excel file can be downloaded

### DIFF
--- a/backend/api/Controllers/STEAController.cs
+++ b/backend/api/Controllers/STEAController.cs
@@ -1,13 +1,12 @@
 
-
-using api.Adapters;
 using api.Dtos;
-using api.Models;
+using api.Excel;
 using api.Services;
+
+using ClosedXML.Excel;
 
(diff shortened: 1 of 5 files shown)


backend/api/Controllers/STEAController.cs
```diff
@@ -1,13 +1,13 @@
...
-using api.Models;
+using api.Excel;
 using api.Services;
 
+using ClosedXML.Excel;
+
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
+
```
eiriksanderfjeld, line 10: Remove whitespace /line breaks

eiriksanderfjeld (approved):

backend/api/Controllers/STEAController.cs
petterwildhagen: Fikser det  i neste PR.

PR #418, opened 2022-03-21
Added components for case name and DG4 date.

DanielBohme:
Also changed case description margin. Updated case view.


fredrikleknes:
Fix the codacy error and youre good! It's the console statement

fredrikleknes (approved):
Approved!

PR #425, opened 2022-03-22
Edit changed to textarea, avoid refresh

fredrikleknes:
When editing description you can now add linebreaks to text and they are kept on save. Site will not refresh on save anymore


petterwildhagen (approved):
Looks good! 
Don't forget to squash before merge.

PR #431, opened 2022-03-24
Surfassetview

ahmedoabdi:
- Added create/update logic on Surf Asset
- Added "Surf" title typography
- Added input field "name" on the asset
- Modified backend to make it possible to send in Surf without cost profile
- Save/submit of Surf should now only be possible if a name is set, and not during import of time series on the cost profile
- Added create button on remaining asset


ahmedoabdi:
- Added create/update logic on Surf Asset
- Added "Surf" title typography
- Added input field "name" on the asset
- Modified backend to make it possible to send in Surf without cost profile
- Save/submit of Surf should now only be possible if a name is set, and not during import of time series on cost profile

eiriksanderfjeld:
Mangler oppdatert av frontend/src/models/types.d.ts

petterwildhagen (approved):
Looks good!!!

PR #432, opened 2022-03-24
Add check for autogenerated models

eiriksanderfjeld:


fredrikleknes (approved):
Look good!
We might need to update the types file again since it was generated previous to the latest commits. But that's fine

PR #443, opened 2022-03-28
Added save button for DG4

DanielBohme:
Switched updateCase (saving) to save button rather than auto save on input. Fixes issue with user getting thrown out of manual input due to updates for every input.


petterwildhagen (approved):
Looks good!!

PR #452, opened 2022-03-30
Disable DG4 save button for different scenarios

DanielBohme:
Disables button when no changes are done. Disables if you don't save on one case and go to another case.


petterwildhagen (approved):
Looks good.

PR #454, opened 2022-03-30
Create WellProject view

taustad:


frontend/src/models/assets/wellproject/WellProject.ts
```diff
@@ -0,0 +1,49 @@
...
+            this.rigMobDemob = data.rigMobDemob ?? 0
+            this.annualWellInterventionCost = data.annualWellInterventionCost ?? 0
+            this.pluggingAndAbandonment = data.pluggingAndAbandonment ?? 0
+        } else {
+            this.id = "00000000-0000-0000-0000-000000000000"
+            this.name = ""
+        }
+    }
+
+    static fromJSON(data: Components.Schemas.WellProjectDto): WellProject {
```
petterwildhagen, line 38: Can be simplified to one method??? They are all the same :)

PR #457, opened 2022-03-30
Remove existing projects from dropdown

fredrikleknes:


frontend/src/Views/CreateProjectView.tsx
```diff
@@ -5,7 +5,7 @@ import {
     TextField,
```
petterwildhagen, line 5: Rename file to ProjectView.tsx ???? The view provides viewing and capability to create projects

petterwildhagen (commented):
Looks good!!
I tested what happened if common lib is down by inserting an exception
image

Only one minor name change CreateProjectView -> ProjectView.

frontend/src/Views/CreateProjectView.tsx
ahmedoabdi: There already exists a ProjectView, since this logic is set in a modal, couldn't he just call it ProjectModal.tsx`?

ahmedoabdi: Or perhaps 'CreateProjectModal'

petterwildhagen (approved):
Looks good.

PR #475, opened 2022-04-04
Feature/472 add name field to drainage strategy

ManveerPandher:
Background for case:
* When creating a new DrainageStrategy, there should be possibility to add a name to the Asset.

What has been done:
* Added a View which is editable according to new, or exisiting name on Drainage Asset.

Why this has to be done:
* So users can add their own name, edit and alter exisiting assets name.


fredrikleknes (approved):
LGTM

PR #485, opened 2022-04-05
DCD-Frontend-Add-Equinor-favicon

eiriksanderfjeld:


ManveerPandher:
Standard way is using an ".ico" extension on the image, but
This works aswell!

ManveerPandher (approved):
LGTM

PR #487, opened 2022-04-05
Delete drainage strategy timeseries

taustad:
- Added drainage timeseries to dbcontext
- Compares new dto to existing drainage model and removes deleted timeseries


petterwildhagen (approved):
Looks good!!!

PR #502, opened 2022-04-06
Fix drainage backend test

taustad:
- Fix drainage strategy backend test

### DIFF
--- a/backend/tests/Services/DrainageStrategyServiceShould.cs
+++ b/backend/tests/Services/DrainageStrategyServiceShould.cs
@@ -149,8 +149,7 @@ namespace tests
             var oldStrategy = CreateTestDrainageStrategy(project);
             fixture.context.DrainageStrategies.Add(oldStrategy);
             fixture.context.SaveChanges();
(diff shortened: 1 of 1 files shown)


backend/tests/Services/DrainageStrategyServiceShould.cs
```diff
@@ -150,6 +150,7 @@ public void UpdateDrainageStrategy()
             fixture.context.DrainageStrategies.Add(oldStrategy);
             fixture.context.SaveChanges();
             var updatedStrategy = CreateUpdatedDrainageStrategy(project);
+            updatedStrategy.Id = oldStrategy.Id;
```
ahmedoabdi, line 153: Would it not make more sense to set ID in CreateUpdateDrainageStrategy()?
Add the line `Id = project.DrainageStrategies.FirstOrDefault(x => x.ProjectId == project.Id).Id`
image

ahmedoabdi: or perhaps more like this:
- add oldStrategy as second parameter and set Id from that`?
image

ahmedoabdi (approved):
LGTM!

PR #504, opened 2022-04-06
Delete substructure timeseries

taustad:
- Delete substructure timeseries
- Set Id on backend update test


backend/api/Context/DcdDbContext.cs
```diff
@@ -19,6 +19,7 @@ public DcdDbContext(DbContextOptions<DcdDbContext> options) : base(options)
         public DbSet<SurfCostProfile>? SurfCostProfile { get; set; }
 
         public DbSet<Substructure>? Substructures { get; set; }
+        public DbSet<SubstructureCostProfile>? substructureCostProfiles { get; set; }
```
ahmedoabdi, line 22: Shouldn't the name be SubstructureCostProfile here? Could you please change that?

backend/tests/Services/SubstructureServiceShould.cs
```diff
@@ -140,7 +140,7 @@ public void UpdateSubstructure()
             fixture.context.SaveChanges();
 
             var updatedSubstructure = CreateUpdatedSubstructure(project);
-
+            updatedSubstructure.Id = oldSubstructure.Id;
```
ahmedoabdi, line 143: Instead of setting it in the update method, can we not just ad Id on the creation of Substructure update object?

ahmedoabdi (changes requested):

ahmedoabdi (approved):
LGTM!

PR #505, opened 2022-04-06
Delete exploration timeseries

taustad:
- Delete exploration timeseries
- Set Id on backend update test


petterwildhagen (approved):
Looks good!!

PR #510, opened 2022-04-06
Added delete of costprofile on Topside asset

ahmedoabdi:
Added delete button with functionality to delete cost profile on Topside
image

### DIFF
--- a/backend/api/Adapters/TopsideAdapter.cs
+++ b/backend/api/Adapters/TopsideAdapter.cs
@@ -19,10 +19,9 @@ namespace api.Adapters
                 ArtificialLift = topsideDto.ArtificialLift,
                 Maturity = topsideDto.Maturity
             };
-            if (topsideDto.CostProfile != null)
-            {
-                topside.CostProfile = Convert(topsideDto.CostProfile, topside);
-            }
+
+            topside.CostProfile = Convert(topsideDto.CostProfile, topside);
+
             return topside;
         }
 
--- a/backend/api/Adapters/TopsideAdapter.cs
+++ b/backend/api/Adapters/TopsideAdapter.cs
@@ -19,10 +19,29 @@ namespace api.Adapters
                 ArtificialLift = topsideDto.ArtificialLift,
                 Maturity = topsideDto.Maturity
             };
-            topside.CostProfile = Convert(topsideDto.CostProfile, topside);
+            if (topsideDto.CostProfile != null)
+            {
+                topside.CostProfile = Convert(topsideDto.CostProfile, topside);
+            }
             return topside;
         }
 
+        public static void ConvertExisting(Topside existing, TopsideDto topsideDto){
+            existing.Id = topsideDto.Id;
+            existing.Name = topsideDto.Name;
+            existing.ProjectId = topsideDto.ProjectId;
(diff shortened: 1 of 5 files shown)


frontend/src/Views/TopsideView.tsx
```diff
@@ -180,7 +193,7 @@ const TopsideView = () => {
                         id="topsideName"
                         name="topsideName"
                         placeholder="Enter topside name"
-                        defaultValue={topside?.name}
+                        value={topside?.name}
```
taustad, line 196: Should value be topsideName instead of "topside?.name"?

backend/api/Adapters/TopsideAdapter.cs
```diff
@@ -19,10 +19,30 @@ public static Topside Convert(TopsideDto topsideDto)
                 ArtificialLift = topsideDto.ArtificialLift,
                 Maturity = topsideDto.Maturity
             };
-            topside.CostProfile = Convert(topsideDto.CostProfile, topside);
+            if (topsideDto.CostProfile != null)
+            {
+                topside.CostProfile = Convert(topsideDto.CostProfile, topside);
+            }
```
taustad, line 25: Is this necessary? There is a check for null in the Convert-method.

frontend/src/Views/TopsideView.tsx
```diff
@@ -170,6 +170,19 @@ const TopsideView = () => {
         }
     }
 
+    const deleteCostProfile = () => {
+        const topsideProjectCopy = new Topside(topside)
+        topsideProjectCopy.costProfile = undefined
```
taustad, line 175: Change "topsideProjectCopy" to "topsideCopy"?

taustad (changes requested):
Check comments

frontend/src/Views/TopsideView.tsx
ahmedoabdi: Good point, makes more sense to have topsideName here.

frontend/src/Views/TopsideView.tsx
ahmedoabdi: Done!

backend/api/Adapters/TopsideAdapter.cs
ahmedoabdi: Not really, removed the null check as it was redundant.

backend/api/Adapters/TopsideAdapter.cs
```diff
@@ -19,10 +19,29 @@ public static Topside Convert(TopsideDto topsideDto)
...
+            existing.OilCapacity = topsideDto.OilCapacity;
+            existing.GasCapacity = topsideDto.GasCapacity;
+            existing.FacilitiesAvailability = topsideDto.FacilitiesAvailability;
+            existing.ArtificialLift = topsideDto.ArtificialLift;
+            existing.Maturity = topsideDto.Maturity;
+
+            if (topsideDto.CostProfile != null)
+            {
+                existing.CostProfile = Convert(topsideDto.CostProfile, existing);
+            }
```
taustad, line 43: Can this null check also be removed?

taustad (changes requested):

taustad (approved):
Looks good

PR #516, opened 2022-04-07
Fix save button for drainage name 

taustad:
- Fix save button for drainage strategy
- Change to value for exploration name field


petterwildhagen (approved):
Looks good!

PR #518, opened 2022-04-07
Adding scripts for EF Migrations

eiriksanderfjeld:


fredrikleknes (commented):
Looks good, apart from a codacy error and lint backend failing. From what i can tell you've ran the 'dotnet ef migrations add <name>' command and pushed the generated code, is that right?

eiriksanderfjeld:
`dotnet ef migrations add` og `dotnet ef database update` is the command. Need to get this into the pipeline as well. @fredrikleknes fixed linting as well.

fredrikleknes (approved):
Looks good!

PR #530, opened 2022-04-08
Console error fix

fredrikleknes:
- Handling of the console error log that is caused by having both defaultvalue and value attributes on input
image


ahmedoabdi (approved):
LGTM!

PR #538, opened 2022-04-12
refactor backend 

ahmedoabdi:
- Refactoring backend where it makes sense

### DIFF
--- a/backend/api/Adapters/CommonLibraryProjectDtoAdapter.cs
+++ b/backend/api/Adapters/CommonLibraryProjectDtoAdapter.cs
@@ -8,8 +8,7 @@ namespace api.Adapters
         public static CommonLibraryProjectDto? Convert(dynamic project)
         {
             var projectDto = new CommonLibraryProjectDto();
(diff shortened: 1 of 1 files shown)


backend/api/Adapters/ProjectDtoAdapter.cs
```diff
@@ -7,19 +7,14 @@ public static class ProjectDtoAdapter
...
-            projectDto.Country = project.Country;
-            projectDto.CreateDate = project.CreateDate;
-            projectDto.ProjectCategory = project.ProjectCategory;
-            projectDto.ProjectPhase = project.ProjectPhase;
-            projectDto.Cases = new List<CaseDto>();
+            var projectDto = ProjectToProjectDto(project);
+
             if (project.Cases != null)
             {
+                if (projectDto.Cases == null)
```
taustad, line 14: Can this ever be null?

ahmedoabdi: No it can not 👍🏿

PR #574, opened 2022-04-27
DCD Frontend: Add DG date input control (starting point)

DanielBohme:
The date picker for every DG date should have a limited date range until the next DG date.

Example DG1 should only be able to pick dates up until DG2 date.

image

image

Starting now with these easy checks until all metadata is finished. Rule discussions will come afterwards.

**This is only a start on rules for DG date input control.** More checks are needed and new cases should be made for these later. The reason for only start rules is due to down prioritizing on this case.


ahmedoabdi (approved):
LGTM!

PR #582, opened 2022-04-28
Set artificial lift in topside model, set facilities availability on case

petterwildhagen:
Add facilities availability to case - this shall NOT be inherited to topside as specified in spreadsheet
Inherit artificial lift to Topside


fredrikleknes (approved):
LGTM

PR #587, opened 2022-04-29
Add remaining metadata to transport

DanielBohme:
Add to Transport view:
Length of gas export pipeline
Length of oil export pipeline
Maturity


fredrikleknes (approved):
Looks good!

PR #589, opened 2022-05-02
Adding logging framework for services in backend. 

eiriksanderfjeld:


backend/api/Services/ProjectService.cs
```diff
@@ -54,18 +60,26 @@ public ProjectDto CreateProject(Project project)
             project.WellProjects = new List<WellProject>();
             project.Explorations = new List<Exploration>();
 
+            Activity.Current?.AddBaggage(nameof(project), JsonConvert.SerializeObject(project, Formatting.None,
```
petterwildhagen, line 63: I uncommented this and both the tests and application runs. This can be deleted?

petterwildhagen (changes requested):
Delete all occurences of Activity.Current?.AddBaggage(nameof(_context.Projects), JsonConvert.SerializeObject(_context.Projects, Formatting.None,
                        new JsonSerializerSettings()
                        {
                            ReferenceLoopHandling = ReferenceLoopHandling.Ignore
                        }));

Otherwise OK. This PR seems to be about getting the framework up and running, we could do a session in the whole team to look at what we should be logging????

backend/api/Services/ProjectService.cs
eiriksanderfjeld: Its used to add object to request to Application Insight to enrich the logging with the actual object.

petterwildhagen (approved):
LGTM after clarifications with Eirik

PR #614, opened 2022-05-04
Remove rig mob demob from case. Set editable on well project

DanielBohme:


fredrikleknes (changes requested):
RigMobDemob should also be removed from backend and frontend models:
* Case.cs
* CaseDto.cs
* CaseAdapter.cs
* CaseDtoAdapter.cs
* Case.ts
* types.d.ts (fixed by running autogenerate-models after removed in above files)

fredrikleknes (approved):
LGTM!

PR #617, opened 2022-05-04
Fixed surf according to latest requests from Atle

fredrikleknes:
* Removed riser count and template count from case
* Removed riser count and template count inheritance from case on surfview
* Added producer count, gas injector count and water injector count to surf model and surfview


petterwildhagen (dismissed):
LGTM

petterwildhagen (approved):

PR #625, opened 2022-05-05
Use new TimeSeries for all assets. 

taustad:
Changes:
- Frontend services for assets implements IAssetService. Move IAssetService from Save.tsx to IAssetService.ts
- Frontend time series implements ITimeSeries
- Use new TimeSeries component in all assets
- Assets implement IAsset interface


petterwildhagen (approved):
LGTM

PR #636, opened 2022-05-10
DCD frontend: Add production profile units to drainage strategy

DanielBohme:
Added units from user story. Having "Oilfield" as placeholder unit for Oil field as we do not have the corresponding units yet.


taustad (approved):
LGTM

PR #637, opened 2022-05-10
Currency override on asset

taustad:
- Add currency to asset models
- Add dropdown to change currency on asset level
- Set currency from project on asset creation


DanielBohme (approved):
LGTM

PR #668, opened 2022-06-08
Make DG3 and DG4 overwritable on facility assets

DanielBohme:


petterwildhagen (approved):
Looks good except one bug which is that when you edit the year on DG3, DG4 date on the assets by typing on the key board the cursor jumps out of the text box ...... we had a similar bug with DG4 date on case back in march.

DanielBohme:
Will add bug as a new task as this is not critical and need to get this out in main.

PR #673, opened 2022-06-10
Add back lowest date check for asset DG3 and DG4

DanielBohme:
Need both null check and lowest date check due to dev environment showing that asset DG3 and DG4 are saved with standard lowest possible date.


petterwildhagen (approved):
LGTM

PR #675, opened 2022-06-13
Assetcurrency bug

fredrikleknes:
There was a mismatch in backend vs frontend on asset currency. Backend: 1 == NOK, 2 == USD. Frontend: 0 == USD, 1 == NOK


ahmedoabdi (approved):
LGTM!

PR #678, opened 2022-06-20
Added FacilityOpex to Topside and Prosp import

DanielBohme:


backend/api/Services/ImportProspService.cs
```diff
@@ -164,6 +164,7 @@ private void ImportTopside(List<Cell> cellData, Guid sourceCaseId, Guid projectI
             var cO2OnMaxOilProfile = ReadDoubleValue(cellData, "L99");
             var cO2OnMaxGasProfile = ReadDoubleValue(cellData, "L100");
             var cO2OnMaxWaterInjectionProfile = ReadDoubleValue(cellData, "L101");
+            var facilityOpex = ReadDoubleValue(cellData, "L80");
```
eiriksanderfjeld, line 167: Should all cell references be moved out to the config? One array for prosp config, to simplify changes without the need to redeploy? @ahmedoabdi @DanielBohme

eiriksanderfjeld (approved):

backend/api/Services/ImportProspService.cs
DanielBohme: Worth a discussion. Not sure how this would apply later to user being able to change fields in admin module.

PR #679, opened 2022-06-20
Timeseries dates alignment when importing from PROSP

ahmedoabdi:
The timeseries date are set based on DG4 date from case, when importing from PROSP the date needs to be changed the value imported from PROSP. A check needs to be done based on the source of data on case and asset.


taustad (approved):
LGTM 👀

PR #680, opened 2022-06-20
Adding file transform step to db migrations

eiriksanderfjeld:


pipelines/templates/stage-deploy-fusion.yml
```diff
@@ -157,3 +162,5 @@ stages:
                       custom: "ef"
                       arguments: "database update --connection \"${{ parameters.DBContext }}\" --verbose"
                       workingDirectory: "$(Build.SourcesDirectory)/backend/api"
+
```
DanielBohme, line 165: Unecessary extra lines?

PR #688, opened 2022-06-23
Add gray line-breaks within CaseView, and align them equally

ManveerPandher:
Reason for case:
* Under the view for CaseView, there was alot of Assets and components tightly visually coupled, with my opinion, making it difficult to separate them.

Azure Task:
* https://dev.azure.com/2S-IAF/DCD/_workitems/edit/274


ahmedoabdi (approved):
LGTM!

PR #692, opened 2022-06-29
Properties missing from create method

eiriksanderfjeld:
Added gas and oil to CreateTransport method for Transport


ahmedoabdi (approved):
LGTM!

PR #695, opened 2022-06-29
Well/ well type correction

DanielBohme:
There should only be a single well in a project. In that well, there should be a list of well types and exploration well types on that project well.

Need to change Wells on case to be well types and exploration well types.
Well types and exploration well types on case will be inherited to Well project and Exploration


ahmedoabdi:
Redundant

PR #697, opened 2022-06-30
SharePoint implementation to support retrieval of PROSP 

ahmedoabdi:
- Use Graph API to retrieve content from SharePoint
- Use [OBO](https://docs.microsoft.com/en-us/graph/sdks/choose-authentication-providers?tabs=CS) auth provider to authenticate user access to SharePoint Site
- Retrieve the list and content of files with types .xlsx and .xlsm


backend/api/Services/GraphRestService.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        public List<Stream> GetAllFilesFromSite()
+        {
+            // var graphClient = new GraphServiceClient(BaseUrl, authenticationProvider);
+            // var driveItem =  graphClient.Me.Drive.Root.ListItem.Request().GetAsync();
+            var graphClient = GetGraphClient();
+            var stream = graphClient.Sites[SiteId].Drive.Items
+                .Request().GetAsync().GetAwaiter().GetResult();
+
+            var files = (stream.Where(item => item != null).Select(item => item.Content)).ToList();
```
ahmedoabdi, line 70: Need to add some sort of type check, and return list with name and content to frontend.

backend/api/Controllers/UploadController.cs
```diff
@@ -3,18 +3,30 @@
...
+
+        public UploadController(ImportProspService prospService, GraphRestService graphRestService)
         {
             _prospService = prospService;
+            _graphRestService = graphRestService;
+        }
+
+        [HttpGet(Name = "GetFilesFromSharePoint")]
+        public List<Stream> GetFilesFromSharePoint()
+        {
```
ahmedoabdi, line 27: Not doing much here yet, but the thought was to send a list of all files from site and let the user decide which PROSP file to load.

backend/api/Controllers/UploadController.cs
```diff
@@ -1,65 +1,52 @@
...
+    {
+        return _graphRestService.GetFilesFromSite();
+    }
 
-                };
+    [HttpPost(Name = nameof(Upload))]
+    [DisableRequestSizeLimit]
+    public ProjectDto? Upload([FromQuery] Guid projectId, [FromQuery] Guid sourceCaseId)
+    {
+        var graph = _graphRestService.GetFilesFromSite();
```
eiriksanderfjeld, line 34: Seems to be able to remove this?

eiriksanderfjeld (approved):

PR #698, opened 2022-07-01
Generate cessation cost profile based on production

taustad:


ahmedoabdi:
Small issue: Creating new well should not be possible before creating the asset(that includes both wellproject and exploration), as the assetId are used as foreign key on well.

ahmedoabdi (approved):
LGTM!

PR #723, opened 2022-08-11
Duplicate case from project overview

DanielBohme:
image

Able to duplicate case.

Currently table only updates after site refresh


frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,6 +175,17 @@ const CasesTableView = ({
             setTableData(sorted)
         }
     }
+
+    const duplicateCase = async () => {
+        try {
+            if (caseRowDataSelected != null) {
+                const newProject: Project = await (await GetCaseService()).duplicateCase(caseRowDataSelected.id, {})
```
taustad, line 182: Should you not set the new project after retrieving it from the server?

taustad: Why specify the type when duplicateCase returns a Promise<Project>?

frontend/src/Services/CaseService.ts
```diff
@@ -15,6 +16,11 @@ class __CaseService extends __BaseService {
         const res: Components.Schemas.ProjectDto = await this.put("", { body })
         return Project.fromJSON(res)
     }
+
+    public async duplicateCase(copyCaseId: any, data: Components.Schemas.CaseDto): Promise<Project> {
```
taustad, line 20: Why datatype any for copyCaseId?

frontend/src/Views/CasesTableView.tsx
```diff
@@ -158,6 +161,8 @@ const CasesTableView = ({
     const [element, setElement] = useState<HTMLButtonElement>()
 
     const [caseRowDataSelected, setCaseRowDataSelected] = useState<any>()
+    const [, setProject] = useState<Project>()
```
taustad, line 164: Pass the setProject function from the view in order for the component to rerender when you duplicate a case.

frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,6 +179,19 @@ const CasesTableView = ({
             setTableData(sorted)
         }
     }
+
+    const duplicateCase = async () => {
+        try {
+            if (caseRowDataSelected != null) {
+                const projectResult = await (await GetProjectService()).getProjectByID(fusionProjectId!)
```
taustad, line 186: What is the purpose of this?

taustad (changes requested):
See comments

frontend/src/Services/CaseService.ts
DanielBohme: datatype is any due to the object: 
image

taustad: Isn't copyCaseId just the id of the case?

backend/api/Services/CaseService.cs
```diff
@@ -44,6 +44,16 @@ public ProjectDto DuplicateCase(Guid caseId)
             }
             var project = _projectService.GetProject(case_.ProjectId);
             case_.Project = project;
+
+            List<Case> duplicateCaseNames = new List<Case>();
+            foreach (Case c in project.Cases!)
+            {
+                if (c.Name.Contains(case_.Name))
```
taustad, line 51: Not a serious bug, but if you have a case named "Test" and a case named "T", and you duplicate "T" it will be named "T - copy #2" and not "T - copy #1"

Here is an example with more values
Capture123123

taustad: Check for equality and not contains

taustad (changes requested):
See comment

backend/api/Services/CaseService.cs
DanielBohme: Changed to equality even though it looks ugly. Not able to replicate "T" name copy bug after equality implementation.

backend/api/Services/CaseService.cs
```diff
@@ -44,6 +44,23 @@ public ProjectDto DuplicateCase(Guid caseId)
...
             case_.Project = project;
+
+            List<Case> duplicateCaseNames = new List<Case>();
+            foreach (Case c in project.Cases!)
+            {
+                if (c.Name.Equals(case_.Name))
+                {
+                    duplicateCaseNames.Add(c);
+                }
+                string lastCharInName = c.Name.Substring(c.Name.Length - 1, 1);
```
taustad, line 55: What happens if there are 10 copies or more?

backend/api/Services/CaseService.cs
```diff
@@ -44,6 +44,23 @@ public ProjectDto DuplicateCase(Guid caseId)
...
+                int copyNumber;
+                var isNum = int.TryParse(lastCharInName, out copyNumber);
```
taustad, line 57: Where is "isNum" used?

backend/api/Services/CaseService.cs
taustad: Why are some variables declared with type some with var?

backend/api/Services/CaseService.cs
DanielBohme: Woops. This isn't the correct version. Will update to correct one

backend/api/Services/CaseService.cs
DanielBohme: That is true. If copies go beyond 10, the next copies beyond that will all be named copy nr 10. Will have to check how much needs to be put into this. If beyond 10 copies is relevant

taustad (approved):
LGTM

PR #728, opened 2022-08-15
Case table - open case, edit case, removed top line in side menu

DanielBohme:
Case can now be opened and edited in the case table menu.
Top line in side menu has been removed.

Currently table only updates after site refresh


frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,67 +197,142 @@ const CasesTableView = ({
...
+    const handleDescriptionChange: ChangeEventHandler<HTMLInputElement> = (e) => {
+        const { value } = e.target
+        setCaseDescription(value)
+    }
+
+    const submitEditCaseForm: MouseEventHandler<HTMLButtonElement> = async (e) => {
+        e.preventDefault()
+
+        try {
+            const projectResult: Project = await (await GetCaseService()).updateCase({
```
taustad, line 215: Why not set the project after updating?

frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,67 +197,142 @@ const CasesTableView = ({
...
+                description: caseDescription,
+                name: caseName,
+                projectId: project.projectId,
+                id: caseRowDataSelected.id,
+            })
```
taustad, line 220: Why create a new Case object instead of spreading the existing one? If you do it this way you lose all the data other than what you specify.

frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,67 +197,142 @@ const CasesTableView = ({
...
+                    </Menu.Item>
+                </Menu>
+            </div>
+            <ModalNoFocus isOpen={editCaseModalIsOpen} title={`Edit case name and description for ${caseRowDataSelected?.name}`}>
+                <EditCaseForm>
+                    <TextField
+                        label="Name"
+                        id="name"
+                        name="name"
+                        placeholder={caseRowDataSelected?.name ?? "Enter a name"}
```
taustad, line 305: Should we not keep the existing value for each case, so that we don't have to add both name and description when chaning name?

frontend/src/Views/CasesTableView.tsx
```diff
@@ -3,14 +3,30 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
```
taustad, line 3: Try to remove unnecessary eslint-disables, or make it so they aren't necessary

taustad (changes requested):
See comments

frontend/src/Views/CasesTableView.tsx
```diff
@@ -147,9 +163,9 @@ interface CasesTableViewProps {
     project: Project
 }
 
-const CasesTableView = ({
+function CasesTableView({
```
taustad, line 166: Why change const to function?

DanielBohme: Was to have uniformity, but I see now that different "view" files use either const or function. Reverted back to const

frontend/src/Views/CasesTableView.tsx
```diff
@@ -174,67 +197,142 @@ const CasesTableView = ({
...
+    }
+
+    const submitEditCaseForm: MouseEventHandler<HTMLButtonElement> = async (e) => {
+        e.preventDefault()
+
+        try {
+            const caseResult = unwrapCase(project.cases.find((o) => o.id === caseRowDataSelected.id))
+            caseResult.name = (caseName === "" ? caseRowDataSelected.name : caseName)
+            caseResult.description = (caseDescription === "" ? caseRowDataSelected.description : caseDescription)
+            const projectResult: Project = await (await GetCaseService()).updateCase(caseResult)
```
taustad, line 218: Why specify the type for projectResult?

taustad (approved):
LGTM

PR #732, opened 2022-08-18
Duplicate case test name check fix

DanielBohme:
Changed name check to description check due to duplicates getting new name.


taustad (approved):
LGTM

PR #742, opened 2022-08-30
Join tables to a single ag-grid table per view

DanielBohme:


frontend/src/Components/TimeSeries.tsx
```diff
@@ -4,145 +4,217 @@ import {
...
-                <DeleteButton
-                    disabled={timeSeries === undefined}
-                    color="danger"
-                    onClick={deleteTimeseries}
-                >
-                    Delete
-                </DeleteButton>
-            </Wrapper>
+            </WrapperTablePeriod>
+            <Typography variant="h2">Cost</Typography>
```
taustad, line 204: The title of the table should probably be a parameter

frontend/src/Components/DataTable/DataTable.tsx
```diff
@@ -1,66 +1,178 @@
...
+    const setUnit = (j:number) => {
+        if (["CO2 emissions", "Production profile NGL"].includes(profileName[j])) {
+            return "MTPA"
+        }
+        if (["Net sales gas", "Fuel flaring and losses", "Production profile gas"].includes(profileName[j])) {
+            return GSM3Enum[profileEnum]
+        }
+        if (["Production profile oil", "Production profile water",
+            "Production profile water injection"].includes(profileName[j])) {
+            return MSM3Enum[profileEnum]
```
taustad, line 62: Is it not better if the view controls the unit of the time series?

frontend/src/Components/TimeSeries.tsx
```diff
@@ -4,145 +4,217 @@ import {
...
-            setGridData([[]])
         }
+        setGridData(combinedTimeseries)
     }
 
     useEffect(() => {
-        buildAlignedGrid(timeSeries!)
+        buildAlignedGrid(combinedTimeseries!)
+
+        if (gridData !== undefined && firstYear?.toString().length === 4 && lastYear?.toString().length === 4
```
taustad, line 87: Make a function that checks for the length of a number instead of calling "firstYear?.toString().length === 4" 8 times.

frontend/src/Components/TimeSeries.tsx
```diff
@@ -4,145 +4,217 @@ import {
...
-        newTimeSeries.startYear = year
-        newTimeSeries.values = input.replace(/(\r\n|\n|\r)/gm, "").split("\t").map((i) => parseFloat(i))
-        setTimeSeries(newTimeSeries)
-        if ((Number(year)
-        + Number(dG4Year!)) < (firstYear ?? Number.MAX_SAFE_INTEGER)) {
-            setFirstYear((Number(year) + Number(dG4Year!)))
+    const createNewGridWithData = (j: any) => {
+        const newTimeSeries: ITimeSeries = { ...timeSeries![j] }
+        const colYears = []
+        for (let c = beginningYear; c! <= endingYear!; c! += 1) {
```
taustad, line 123: Either properly check for null / undefined or set default values instead of using ! everywhere

frontend/src/Components/TimeSeries.tsx
```diff
@@ -4,145 +4,217 @@ import {
...
-    setFirstYear,
-    setLastYear,
+    profileName,
+    profileEnum,
 }: Props) => {
     const [columns, setColumns] = useState<string[]>([""])
     const [gridData, setGridData] = useState<CellValue[][]>([[]])
-    const [dialogOpen, setDialogOpen] = useState(false)
+    const [beginningYear, setBeginningYear] = useState<number | undefined>()
+    const [endingYear, setEndingYear] = useState<number | undefined>()
```
taustad, line 39: Could this have a better name? Could you make the difference between firstYear and beginningYear more clear?

frontend/src/Components/TimeSeries.tsx
```diff
@@ -4,145 +4,217 @@ import {
...
-            + Number(dG4Year!) + Number(newTimeSeries.values.length))
+
+        if (endingYear! < (Number(colYears[0]) + Number(timeSeries![j].values!.length))) {
+            const yearDifference = (Number(colYears[0]) + timeSeries![j].values!.length - 1) - endingYear!
+            newTimeSeries.values = timeSeries![j].values?.slice(0, -yearDifference)
         }
-        buildAlignedGrid(newTimeSeries)
-        setDialogOpen(!dialogOpen)
+
+        if (beginningYear! > (Number(timeSeries![j].startYear!) + Number(dG4Year!))) {
```
taustad, line 153: Make funcitons of the conditions in the if statements so that the user doesn't have to compute the expression every time. 

Instead of "beginningYear! > (Number(timeSeries![j].startYear!) + Number(dG4Year!)))" make a function called TableYearGreaterThanProfileYear() or similar

frontend/src/Components/DataTable/DataTable.tsx
```diff
@@ -1,66 +1,178 @@
...
+                }
+                const rowObj = objKey
+                    .reduce((obj:any, element:any, index:any) => ({ ...obj, [element]: objVal[index] }), {})
+                combinedObjArr.push(rowObj)
+
+                const totalCostObj = { "Total cost": totalCost }
+                rowTotalCost.push({ ...combinedObjArr[j], ...totalCostObj, ...rowPinned })
+            }
+        }
+        return rowTotalCost
```
taustad, line 110: Change cost to value

frontend/src/Components/TimeSeries.tsx
taustad: Create a function that takes a number as a parameter and returns whether or not the length is 4.
Don't add new variables that sotre the length of other numbers

frontend/src/Components/DataTable/DataTable.tsx
taustad: In the whole component, not just 110

taustad (approved):

PR #759, opened 2022-09-13
Formating updates

eiriksanderfjeld:
Removed commonlib references, updated formating and enabled ShowPII to
devenv.


taustad (approved):
LGTM

PR #768, opened 2022-09-22
PROSP unlinking of assets when changing files

ahmedoabdi:
During a change of prosp file in the technical input menu, assets checked as imported from the previous file will be removed from the case ( or more accurately unlinked from the case).


ahmedoabdi (commented):
LGTM!

taustad (approved):

taustad (approved):

PR #773, opened 2022-09-26
Join readonly profiles to merged profile tables

DanielBohme:


ahmedoabdi:
Apply button throws error and routing on exploration and wellproject fails when creating new

PR #790, opened 2022-10-19
Production profiles tab fixes, add cost and summary tab

taustad:
- Set update and modify time on case only on backend
- Add NPV, BE, feed studies and feasibility to case model
- Remove facilities availability from topside and drainage strategy
- Add updated endpoints to wellproject and exploration that returns the entity
- Add gas solution to drainage strategy
- Remove number in case name when duplicating
- Use loading icon when waiting for server when clicking Create / Save button on edit / new case modal 
- Add case cost tab
- Set table max and min year
- Add case summary tab
- Make case view header visible when scrolling tab view


ahmedoabdi (changes requested):

ahmedoabdi (approved):
LGTM!

PR #800, opened 2022-10-25
Case functions in "three dot menu" in CaseView

DanielBohme:


frontend/src/Views/CaseView.tsx
```diff
@@ -146,6 +146,8 @@ const CaseView = () => {
```
ahmedoabdi, line 146: Remove import of 'CreateCaseModal' on line 38, and align-content on line 104 has invalid value check: https://developer.mozilla.org/en-US/docs/Web/CSS/align-content

ahmedoabdi (commented):
Still not redirecting to overview after renaming case,

ahmedoabdi (approved):
LGTM!

PR #812, opened 2022-10-27
Align drilling schedule tables

taustad:
- Facilities availability show percentage in frontend, store as 0 - 1.0
- Align drilling schedules
- Set table years from drilling schedules
- Fix navigation when navigating from a case to another project


DanielBohme (approved):
LGTM

PR #816, opened 2022-10-27
Case table pinned columns

DanielBohme:
Also added decimal rounding to row totals which I forgot to add earlier.
Added pinned columns to drilling schedule and row totals


frontend/src/Views/Case/CaseDrillingScheduleTabTable.tsx
```diff
@@ -91,8 +92,13 @@ function CaseDrillingScheduleTabTable({
...
                     for (let i = tableWell.drillingSchedule.startYear;
                         i < tableWell.drillingSchedule.startYear + tableWell.drillingSchedule.values.length; i += 1) {
-                        tableWell[(dg4Year + i).toString()] = tableWell.drillingSchedule.values[j]
+                        tableWell[(dg4Year + i).toString()] = tableWell.drillingSchedule.values.map(
+                            (v:number) => Math.round((v + Number.EPSILON) * 10) / 10,
+                        )[j]
                         j += 1
+                        tableWell.total = tableWell.drillingSchedule.values.map(
+                            (v:number) => Math.round((v + Number.EPSILON) * 10) / 10,
+                        ).reduce((x: number, y: number) => x + y)
```
taustad, line 101: I think rounding here is unnecessary. Drilling schedule should be integer values, so I think we need to ensure input is a whole number instead of rounding to one decimal.

taustad (approved):

PR #818, opened 2022-10-28
Add CO2 tab, remove asset dropdowns, add lock to generated timeseries

taustad:
- Add CO2 tab
- Add well count for different categories to drilling schedule tab
- Add lock icon to readonly rows in table
- Remove asset dropdown in CaseView


ahmedoabdi (approved):
LGTM!

PR #819, opened 2022-10-28
Add loading icon when fetching fusion config

taustad:
Add loading icon when opening case
Open description tab when changing case


DanielBohme (approved):
LGTM

PR #834, opened 2022-10-31
Metadata to co2 tab

DanielBohme:


frontend/src/Views/Case/CaseCO2Tab.tsx
```diff
@@ -72,6 +88,8 @@ function CaseCO2Tab({
     const [endYear, setEndYear] = useState<number>(2030)
     const [tableYears, setTableYears] = useState<[number, number]>([2020, 2030])
 
+    const [isSaving, setIsSaving] = useState<boolean>()
```
taustad, line 91: What is this variable used for?

taustad (approved):

taustad (approved):

PR #849, opened 2022-11-03
Add copy functionality to assets

taustad:


backend/api/Services/WellProjectWellService.cs
```diff
@@ -102,6 +102,28 @@ public WellProjectWell GetWellProjectWell(Guid wellId, Guid caseId)
         return wellProjectWell;
     }
 
+    public WellProjectWellDto[]? CopyWellProjectWell(Guid sourceWellProjectId, Guid targetWellProjectId)
+    {
+        var sourceWellProjectWells = GetAll().Where(wpw => wpw.WellProjectId == sourceWellProjectId).ToList();
+        if (sourceWellProjectWells?.Count > 0)
```
ahmedoabdi, line 108: Codacy warning, perhaps inverting the if-check in line 108 would handle the warning: 
 if (!(sourceWellProjectWells?.Count > 0))
        {
            return null;
        }`

ahmedoabdi (approved):

PR #870, opened 2022-11-11
Prosp import case auto select

DanielBohme:
Auto select case when a change has been performed in the table


eiriksanderfjeld (approved):

ahmedoabdi (approved):
LGTM!

PR #879, opened 2022-11-17
Reduce database calls in study, opex and cessation generation

taustad:
- Reduce database calls in study, opex and cessation generation
- Use values from operational well costs in cessation and opex calculation


ahmedoabdi (approved):
LGTM!

PR #883, opened 2022-11-21
Compare cases tab with charts

DanielBohme:


backend/api/Services/CompareCasesService.cs
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+        catch (ArgumentException)
+        {
+            _logger.LogInformation("DrainageStrategy {0} not found.", caseItem.DrainageStrategyLink);
+        }
+
+        if (project.PhysicalUnit != 0 && !excludeOilFieldConversion)
+        {
+            return sumGasProduction * 35.315 / 1E9;
```
ahmedoabdi, line 129: Too many magical numbers, please do use variables instead

ahmedoabdi (changes requested):
Update service with variables instead of magic numbers

ahmedoabdi (approved):
LGTM!

PR #894, opened 2022-11-29
Refresh project from backgroundservice

eiriksanderfjeld:
Update project if it differs from Project Master, running once each day

### DIFF
--- a/backend/api/Services/ProjectService.cs
+++ b/backend/api/Services/ProjectService.cs
@@ -283,21 +283,30 @@ public class ProjectService
 
     private ProjectDto GetProjectDtoFromProjectMaster(Guid projectGuid)
     {
-        var projectMaster = _fusionService.ProjectMasterAsync(projectGuid).GetAwaiter().GetResult();
-        var category = CommonLibraryProjectDtoAdapter.ConvertCategory(projectMaster.ProjectCategory ?? "");
-        var phase = CommonLibraryProjectDtoAdapter.ConvertPhase(projectMaster.Phase ?? "");
-        ProjectDto projectDto = new()
+        if (_fusionService != null)
         {
-            Name = projectMaster.Description ?? "",
-            CommonLibraryName = projectMaster.Description ?? "",
-            FusionProjectId = projectMaster.Identity,
-            Country = projectMaster.Country ?? "",
-            Currency = Currency.NOK,
-            PhysUnit = PhysUnit.SI,
-            ProjectId = projectMaster.Identity,
-            ProjectCategory = category,
-            ProjectPhase = phase,
-        };
-        return projectDto;
+            var projectMaster = _fusionService.ProjectMasterAsync(projectGuid).GetAwaiter().GetResult();
+            var category = CommonLibraryProjectDtoAdapter.ConvertCategory(projectMaster.ProjectCategory ?? "");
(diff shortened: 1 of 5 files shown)


backend/api/Services/ProjectService.cs
```diff
@@ -244,4 +273,48 @@ private Project AddAssetsToProject(Project project)
         project.Wells = _wellService.GetWells(project.Id).ToList();
         return project;
     }
+
+    public void UpdateProjectFromProjectMaster()
+    {
+        var projectsDtos = GetAllDtos();
+        int numberofdeviations = 0;
+        int totalprojects = projectsDtos.Count();
```
taustad, line 281: Use camel case

backend/api/Services/ProjectService.cs
```diff
@@ -244,4 +273,48 @@ private Project AddAssetsToProject(Project project)
         project.Wells = _wellService.GetWells(project.Id).ToList();
         return project;
     }
+
+    public void UpdateProjectFromProjectMaster()
+    {
+        var projectsDtos = GetAllDtos();
+        int numberofdeviations = 0;
```
taustad, line 280: Any reason to use int instead of var?

backend/api/Services/ProjectService.cs
```diff
@@ -244,4 +273,48 @@ private Project AddAssetsToProject(Project project)
...
+
+    private ProjectDto GetProjectDtoFromProjectMaster(Guid projectGuid)
+    {
+        var projectMaster = _fusionService.ProjectMasterAsync(projectGuid).GetAwaiter().GetResult();
+        var category = CommonLibraryProjectDtoAdapter.ConvertCategory(projectMaster.ProjectCategory ?? "");
+        var phase = CommonLibraryProjectDtoAdapter.ConvertPhase(projectMaster.Phase ?? "");
+        ProjectDto projectDto = new()
+        {
+            Name = projectMaster.Description ?? "",
+            Description = projectMaster.Description ?? "",
```
taustad, line 308: Seems like project name and description both use the same field from project master. Should we create another field for user input description for project or use the existing description property?

backend/api/Services/ProjectService.cs
```diff
@@ -244,4 +273,48 @@ private Project AddAssetsToProject(Project project)
...
+                _logger.LogInformation("Project {projectName} ({projectId}) is identical to ProjectMaster", project.Name, project.ProjectId);
+            }
+        }
+        _logger.LogInformation("Number of projects which differs from ProjectMaster: {count} / {total}", numberofdeviations, totalprojects);
+    }
+
+
+    private ProjectDto GetProjectDtoFromProjectMaster(Guid projectGuid)
+    {
+        var projectMaster = _fusionService.ProjectMasterAsync(projectGuid).GetAwaiter().GetResult();
```
taustad, line 302: Add null check for _fusionService if ProjectService is created with the wrong constructor

backend/api/Services/ProjectService.cs
```diff
@@ -244,4 +273,48 @@ private Project AddAssetsToProject(Project project)
         project.Wells = _wellService.GetWells(project.Id).ToList();
         return project;
     }
+
+    public void UpdateProjectFromProjectMaster()
+    {
+        var projectsDtos = GetAllDtos();
```
taustad, line 279: projectDtos

taustad (changes requested):

taustad (approved):

PR #905, opened 2022-12-06
Divide each power use calculation by facilities availability

taustad:
Use drilling schedule start year in co2emissions


DanielBohme (approved):
LGTM

PR #907, opened 2022-12-08
Reference case 

ahmedoabdi:
- Only one reference case on project can be set
- Now set it on Project as GUID
- Add icon that specifies it as Reference Case


DanielBohme (approved):
LGTM

PR #927, opened 2023-01-16
Gas solution export injection handling

DanielBohme:
Set Net sales gas to empty if gas solution is injection


ahmedoabdi (commented):
`if (drainageStrategy.GasSolution > 0)` instead?

PR #928, opened 2023-01-17
Reference case visual

DanielBohme:


frontend/src/Components/SideMenu/ProjectMenuItemComponent.tsx
```diff
@@ -59,6 +64,13 @@ function ProjectMenuItemComponent({ item, projectId, subItems }: Props) {
         setIsOpen(isSelected)
     }, [isSelected])
 
+    const withReferenceCase = (c: any) => {
```
taustad, line 67: Any reason to have this function here instead of in MenuItem?

DanielBohme: moved to menuitem

taustad (approved):

PR #1044, opened 2024-02-29
Misc fixes

taustad:
Fix generating profiles 
Fix setting created date for case
Fix setting default DG4 date on case creation


JonasABach (approved):
LGTM

PR #1114, opened 2024-05-03
feat: Add endpoints for updating DrainageStrategy, Topside and Substructure

taustad:
- Renamed instances of namespace Api to api
- Enforced LF line endings in .editorconfig
- Moved services to appropriate folders
- Added services for mapping to DTOs
- Added endpoints to update assets and timeseries
- Added services to update assets and timeseries
- Added repositories to handle data access
- Added custom exception for mapping error


JonasABach (approved):
Lgtm

PR #1124, opened 2024-05-21
feat: Add DeferredOilProduction and DeferredGasProduction to production profiles

DanielBohme:
Added DeferredOilProduction and DeferredGasProduction to model and services.
Added hide profiles without values functionality when table is in read only. Profiles will still appear when in edit mode.


arian-garshi (approved):
lgtm

PR #1135, opened 2024-05-29
Backend store images

JonasABach:
implemented so that users can upload images to webapp which will store the image in azure blob storage and the image URL in the database


frontend/src/Services/ImageUploadService.tsx
```diff
@@ -0,0 +1,29 @@
+import { __BaseService } from "./__BaseService"
+import { config } from "./config"
+import { getToken, loginAccessTokenKey } from "../Utils/common"
+
+export class ImageUploadService extends __BaseService {
```
taustad, line 5: Rename to ImageService? This service both gets and posts images

package.json
```diff
@@ -1,9 +1,15 @@
...
   "lint-staged": {
-      "*.{j,t}s?(x)": "npx eslint --fix"
+    "*.{j,t}s?(x)": "npx eslint --fix"
   },
   "devDependencies": {
     "husky": "^7.0.4",
     "lint-staged": "^12.3.4"
+  },
+  "dependencies": {
+    "@azure/storage-blob": "^12.18.0",
```
taustad, line 10: Are these dependencies needed? They are added to the package.json outside of the frontend folder

backend/api/Context/DcdDbContext.cs
```diff
@@ -93,6 +94,11 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         modelBuilder.Entity<Project>()
             .Property(p => p.Classification)
             .HasDefaultValue(ProjectClassification.Internal);
+
+        modelBuilder.Entity<Case>()
```
taustad, line 98: Is it necessary to specify a one to many relationship in OnModelCreating?

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,59 @@
+using Microsoft.AspNetCore.Mvc;
+using api.Context;
+using Microsoft.EntityFrameworkCore;
+using api.Models;
+using Microsoft.AspNetCore.Authorization;
+
+[Authorize]
+[ApiController]
+[Route("projects/{projectId}/cases/{caseId}/[controller]")]
```
taustad, line 9: Don't use [controller] here, specify the endpoint with a string. I think we should call it "images"

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,59 @@
...
+    private readonly IBlobStorageService _blobStorageService;
+    private readonly DcdDbContext _context;
+
+    public BlobStorageController(IBlobStorageService blobStorageService, DcdDbContext context)
+    {
+        _blobStorageService = blobStorageService;
+        _context = context;
+    }
+
+    [HttpPost("upload")]
```
taustad, line 21: If the path is cases/caseId/images, you can post directly the endpoint, no need for "upload"

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        var imageUrl = await _blobStorageService.UploadImageAsync(imageBytes, contentType, blobName);
+
+        var newImage = new Image
+        {
+            CaseId = caseId,
+            Url = imageUrl,
+        };
+        _context.Images.Add(newImage);
+        await _context.SaveChangesAsync();
```
taustad, line 44: I think most of this should be moved to the _blobStorageService. The data access should be moved to either CaseRepository or create an ImageRepository to read and write to the database

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        return Ok(new { imageUrl });
+    }
+
+    [HttpGet("images")]
+    public async Task<IActionResult> GetImages(Guid caseId)
+    {
+        var images = await _context.Images
```
taustad, line 52: The controller should not use the databasecontext directly. Move this to a method in BlobStorageService

backend/api/Models/Image.cs
```diff
@@ -0,0 +1,21 @@
...
+    [Required]
+    public string Url { get; set; } = string.Empty!;
+    public DateTimeOffset CreateTime { get; set; }
+
+    public string Description { get; set; } = string.Empty!;
+
+    [ForeignKey("Case")]
+    public Guid CaseId { get; set; }
+
+    public Case Case { get; set; } = null!;
```
taustad, line 20: I don't think we need Case as a navigation property

backend/api/Program.cs
```diff
@@ -246,6 +257,12 @@
     });
 });
 
+var azureBlobStorageConnectionString = builder.Configuration["AzureBlobStorageConnectionStringForImageUpload"];
+
+builder.Services.AddSingleton(x => new BlobServiceClient(azureBlobStorageConnectionString));
+builder.Services.AddSingleton<IBlobStorageService>(x =>
```
taustad, line 263: Should these be added as singeltons?

backend/api/Services/BlobStorageService.cs
```diff
@@ -0,0 +1,73 @@
...
+    public Task<string> GetBlobSasUrlAsync(string blobName)
+    {
+        var containerClient = _blobServiceClient.GetBlobContainerClient(_containerName);
+        var blobClient = containerClient.GetBlobClient(blobName);
+
+        var sasBuilder = new BlobSasBuilder
+        {
+            BlobContainerName = _containerName,
+            BlobName = blobName,
+            Resource = "b",
```
taustad, line 26: What is this?

backend/api/Services/BlobStorageService.cs
```diff
@@ -0,0 +1,73 @@
...
+    private readonly BlobServiceClient _blobServiceClient;
+    private readonly string _containerName;
+
+    public BlobStorageService(BlobServiceClient blobServiceClient, string containerName)
+    {
+        _blobServiceClient = blobServiceClient;
+        _containerName = containerName;
+    }
+
+    public Task<string> GetBlobSasUrlAsync(string blobName)
```
taustad, line 17: Is this method used anywhere?

backend/api/Context/DcdDbContext.cs
JonasABach: Kan eg bare fjerna den eller?

frontend/src/Components/Gallery/ImageUpload.tsx
```diff
@@ -1,56 +1,90 @@
...
+}
 
-    & p {
-        text-align: center;
-        opacity: 0.9;
-    }
+& p {
+    text-align: center;
+    opacity: 0.9;
+}
```
taustad, line 37: Why remove the indentation?

frontend/src/Services/ImageUploadService.tsx
```diff
@@ -0,0 +1,29 @@
...
+        const formData = new FormData()
+        formData.append("image", file)
+
+        const response = await this.post(`projects/${projectId}/cases/${caseId}/blob-storage/upload`, {
+            body: formData,
+            headers: {
+                "Content-Type": "multipart/form-data",
+            },
+        })
+        console.log("url", response.imageUrl)
```
taustad, line 16: remove console log

backend/api/Context/DcdDbContext.cs
taustad: I think EF figures out the relationship from the models, but please double check

backend/api/Models/Image.cs
JonasABach: So, remove all?

backend/api/Services/BlobStorageService.cs
JonasABach: Updated, now it is

backend/api/Services/BlobStorageService.cs
JonasABach: what is GenerateSasTokenForBlob or what "Resource = "b","?

backend/api/Context/DcdDbContext.cs
JonasABach: Correct. still seems to work

backend/api/Program.cs
JonasABach: tbh, idk what this does, so idk

backend/api/Models/Image.cs
taustad: Keep the CaseId, but remove Case, unless you need to go from image to case somewhere

backend/api/Services/BlobStorageService.cs
taustad: Yeah, what is the Resource = "b", why "b"?

JonasABach: blob storage

frontend/src/Services/config.ts
```diff
@@ -35,6 +35,9 @@ const configuration = {
     CompareCasesService: {
         BASE_URL: "",
     },
+    ImageService: {
+        BASE_URL: "http://localhost:5000",
```
taustad, line 39: Set the BASE_URL in buildConfig below, so that it works in other environemnts

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        var imageUrl = await _blobStorageService.SaveImageAsync(image);
+
+        var newImage = new Image
+        {
+            CaseId = caseId,
+            Url = imageUrl,
+        };
+        _context.Images.Add(newImage);
```
taustad, line 36: Don't call dbcontext from the controller

backend/api/Controllers/BlobStorageController.cs
```diff
@@ -0,0 +1,47 @@
...
+[Route("projects/{projectId}/cases/{caseId}/images")]
+public class BlobStorageController : ControllerBase
+{
+    private readonly IBlobStorageService _blobStorageService;
+    private readonly DcdDbContext _context;
+
+    public BlobStorageController(IBlobStorageService blobStorageService, DcdDbContext context)
+    {
+        _blobStorageService = blobStorageService;
+        _context = context;
```
taustad, line 18: remove context

backend/api/Models/Image.cs
```diff
@@ -0,0 +1,19 @@
...
+public class Image
+{
+    [Key]
+    public Guid Id { get; set; }
+
+    [Required]
+    public string Url { get; set; } = string.Empty!;
+    public DateTimeOffset CreateTime { get; set; }
+
+    public string Description { get; set; } = string.Empty!;
```
taustad, line 15: Can this be nullable?

backend/api/Program.cs
```diff
@@ -246,6 +257,16 @@
...
 
+var azureBlobStorageConnectionString = builder.Configuration["AzureBlobStorageConnectionStringForImageUpload"];
+
+builder.Services.AddScoped(x => new BlobServiceClient(azureBlobStorageConnectionString));
+
+builder.Services.AddScoped<IBlobStorageService, BlobStorageService>((serviceProvider) =>
+{
+    var blobServiceClient = serviceProvider.GetRequiredService<BlobServiceClient>();
+    var dbContext = serviceProvider.GetRequiredService<DcdDbContext>();
+    return new BlobStorageService(blobServiceClient, "upload", dbContext);
```
taustad, line 268: If you remove containerName as a constructor argument, dependency injection should handle the services. 
I think we should get the containerName from azure config inside the service and not in the service constructor

backend/api/Services/BlobStorageService.cs
```diff
@@ -0,0 +1,113 @@
...
+        await sasBlobClient.UploadAsync(stream, new BlobHttpHeaders { ContentType = contentType });
+
+        var imageUrl = blobClient.Uri.ToString();
+
+        return imageUrl;
+    }
+
+    public async Task<IEnumerable<string>> GetImageUrlsAsync(Guid caseId)
+    {
+        var images = await _context.Images
```
taustad, line 83: Use the ImageRepository to access the dbcontext, remove dbcontext from this service

JonasABach: Should i remove the context from the whole service or just this method?

taustad: from the service. If you want to access the database, use the ImageRepository

taustad (approved):

PR #1139, opened 2024-06-04
Backend store images 2.0

JonasABach:
Improved so that images are stored in the blob storage as /projectname/caseId/imageId


DanielBohme (approved):
LGTM

PR #1149, opened 2024-06-06
Add endpoint and services for updating wells

taustad:
- Add endpoints for updating Wells, WellProjectWell and ExplorationWell
- Move SaveChanges outside of repository methods
- Create BaseRepository and add SaveChanges to be called from services

### DIFF
--- a/backend/api/Repositories/CaseRepository.cs
+++ b/backend/api/Repositories/CaseRepository.cs
@@ -6,138 +6,116 @@ using Microsoft.EntityFrameworkCore;
 
 namespace api.Repositories;
 
-public class CaseRepository : ICaseRepository
+public class CaseRepository : BaseRepository, ICaseRepository
 {
-    private readonly DcdDbContext _context;
     private readonly ILogger<CaseRepository> _logger;
 
     public CaseRepository(
         DcdDbContext context,
         ILogger<CaseRepository> logger
-        )
+        ) : base(context)
     {
-        _context = context;
         _logger = logger;
     }
 
     public async Task<Case?> GetCase(Guid caseId)
     {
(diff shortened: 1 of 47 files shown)


backend/api/Repositories/CaseRepository.cs
```diff
@@ -152,21 +130,21 @@ public async Task UpdateModifyTime(Guid caseId)
...
-        }
-        catch (DbUpdateException ex)
-        {
-            _logger.LogWarning(ex, "An error occurred while updating ModifyDate for the Case with id {caseId}.", caseId);
-        }
-        catch (Exception ex)
-        {
-            _logger.LogWarning(ex, "Failed to update modify time for case id {CaseId}, but operation continues.", caseId);
-        }
+        // try
```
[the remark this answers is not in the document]
taustad: I'll look into moving it to the baserepository / services, but for now I think we will keep it commented out

backend/api/Services/Entities/TransportService.cs
```diff
@@ -77,8 +77,8 @@ public async Task<TransportWithProfilesDto> CopyTransport(Guid transportId, Guid
             newTransportDto.CessationCostProfile.Id = Guid.Empty;
         }
 
-        // var topside = await NewCreateTransport(newTransportDto, sourceCaseId);
-        // var dto = TransportDtoAdapter.Convert(topside);
+        // var transport = await NewCreateTransport(newTransportDto, sourceCaseId);
```
DanielBohme, line 80: are these comments necessary?

taustad: These are old comments, I only renamed the commented out variable through find and replace. The whole method will be reworked once we move away from the old API

PR #1160, opened 2024-06-12
feat: Update timeSeries to use react query

DanielBohme:


frontend/src/Components/Case/Components/ClickableLockIcon.tsx
```diff
@@ -18,18 +19,25 @@ const LockIcon: React.FC<LockIconProps> = ({
     setOverrideProfile,
 }) => {
     const handleLockIconClick = (params: any) => {
+        console.log("params: ", params)
```
taustad, line 22: Do we need to log this?

DanielBohme: nope. unsure why I have this console.log. will fix

arian-garshi (approved):

PR #1164, opened 2024-06-13
Fix aggrid console spam

JonasABach:
fixed error
"Warning: validateDOMNesting(...): "<div cannot appear as a descendant of "<p.
By changing from component Typography to Box. Typography renders a "<p tag by default and should not contain block-level elements like "<div like it did. The Box components renders a "<div by default.

Fixed so the consol isnt spammed with messages when rendering page, both when cells are empty but also populated


DanielBohme (approved):
LGTM

PR #1181, opened 2024-06-24
Calculate entities based on changes to other entities

taustad:


backend/api/Services/Entities/Case/CaseService.cs
```diff
@@ -224,23 +225,25 @@ TDto updatedCaseDto
         var existingCase = await _repository.GetCase(caseId)
             ?? throw new NotFoundInDBException($"Case with id {caseId} not found.");
 
+
         _mapperService.MapToEntity(updatedCaseDto, existingCase, caseId);
 
         existingCase.ModifyTime = DateTimeOffset.UtcNow;
 
-        Case updatedCase;
+        // Case updatedCase;
```
DanielBohme, line 233: should these comments be here or is there something that is not finished?

backend/api/Services/Entities/Case/CaseTimeSeriesService.cs
```diff
@@ -335,12 +335,12 @@ UpdateAdditionalOPEXCostProfileDto updatedCostProfileDto
 
         _mapperService.MapToEntity(updatedCostProfileDto, existingProfile, caseId);
 
-        TProfile updatedProfile;
+        // TProfile updatedProfile;
```
DanielBohme, line 338: should these comments be here or is there something that is not finished?

backend/api/Services/GenerateCostProfiles/NetSaleGasProfileService.cs
```diff
@@ -51,17 +51,18 @@ public async Task<NetSalesGasDto> Generate(Guid caseId)
...
 
         return dto ?? new NetSalesGasDto();
     }
 
-    private async Task<int> UpdateDrainageStrategyAndSave(DrainageStrategy drainageStrategy, NetSalesGas netSalesGas)
+    private void UpdateDrainageStrategyAndSave(DrainageStrategy drainageStrategy, NetSalesGas netSalesGas)
     {
         drainageStrategy.NetSalesGas = netSalesGas;
-        return await _context.SaveChangesAsync();
+        return;
```
DanielBohme, line 64: I see that there is a simple return here but on other files the entire return is commented out

DanielBohme (approved):
LGTM

PR #1260, opened 2024-08-20
Add project access policy

taustad:
- Fix authorization always requiring the default role
- Add an attribute to determine whether an endpoint is edit or read
- Check project from url matches project from resource
- Add Fusion service to return members from orgchart
- Add middleware to set HTTP status code based on exception


JonasABach (approved):
lgtm

PR #1262, opened 2024-08-25
fix: Set correct years in Summary tab and fix settings sidebar navigation

DanielBohme:


frontend/src/Components/Case/Components/CaseTabTableHelper.ts
```diff
@@ -74,3 +74,48 @@ export const SetTableYearsFromProfiles = (
...
+
+export const SetSummaryTableYearsFromProfiles = (
+    profiles: (
+        | {
+            id?: string;
+            startYear?: number;
+            name?: string;
+            values?: number[] | null;
+            sum?: number | undefined;
+        }
```
taustad, line 86: use ITimeSeries

taustad (approved):

PR #1269, opened 2024-09-12
Fix check for using override profile in calculations

taustad:
- Fix check for whether or not to use override profiles in calculations
- Add lazy loading library for EF and set navigation properties as virtual
- Add NoTracking to GetCaseWithAssets


JonasABach (approved):
lgtm

PR #1270, opened 2024-09-12
fix: Navigate to project overview when changing from case view from another project

DanielBohme:


frontend/src/Views/CaseView.tsx
```diff
@@ -49,10 +50,15 @@ const CaseView = () => {
         }
     }, [tab])
 
+    useEffect(() => {
+        if (!project?.cases.find((c: any) => c.id === caseId)) {
```
taustad, line 54: avoid the use of any

taustad (approved):

PR #1278, opened 2024-09-18
Add lock for recalculation

taustad:
- Add lock for running calculations
- Add retry for  DbUpdateConcurrencyException
- Rework CompareCases to use stored timeseries
- Rework calculations to update existing timeseries 
- Added EF Split Queries https://learn.microsoft.com/en-us/ef/core/querying/single-split-queries 
- Make GetAllWells in CostProfileFromDrillingScheduleHelper use IQueryable


DanielBohme (approved):
LGTM

PR #1287, opened 2024-09-26
52733 archived cases

H-Trym:
[52733](https://statoil-proview.visualstudio.com/Roughtechs/_workitems/edit/52733)


frontend/src/Components/Case/OverviewCasesTable/CasesAgGridTable.tsx
```diff
@@ -159,10 +190,11 @@ const CasesAgGridTable = ({
         },
     ])
 
-    const casesToRowData = () => {
+    const casesToRowData = (isArchived: boolean) => {
         if (apiData.cases) {
+            let cases = isArchived ? apiData.cases.filter((c) => c.archived === false) : apiData.cases.filter((c) => c.archived === true)
```
taustad, line 195: Do you need to check with c.archived === false and c.archived === true? Can you not just use c.archived and !c.archived?

taustad (approved):

PR #1291, opened 2024-09-27
Fix: Issues with long case names in sidebar

H-Trym:
[56637](https://statoil-proview.visualstudio.com/Roughtechs/_workitems/edit/56637)


JonasABach (approved):
lgtm

PR #1292, opened 2024-10-01
fix disable add case option from drop down menu on an archived case

H-Trym:
[52733](https://statoil-proview.visualstudio.com/Roughtechs/_workitems/edit/52733/)
Look at comment from Anne


DanielBohme (approved):
lgtm

PR #1305, opened 2024-10-04
Refactor create project

taustad:
Add check when creating project for whether or not a project with the same external id exists.

Remove exception from fusionservice and instead return null if no projects are found or the mapping fails

Move ProspImportSettings to its own namespace


JonasABach (approved):
Lgtm

PR #1312, opened 2024-10-11
feat: Create revision modal

DanielBohme:


H-Trym (changes requested):
Some minor comments here in the code review. But the bigger issue is that when you have created a revision and click "Project revisions" the newly created revision is not there, and the user has to refresh. Also there are some errors in the console related to the create revision modal, see PR teams channel for more

H-Trym (approved):
lgtm 👍

PR #1317, opened 2024-10-16
Access case in revision

JonasABach:


frontend/src/Components/Case/OverviewCasesTable/CasesDropMenu.tsx
```diff
@@ -198,23 +194,6 @@ const CasesDropMenu = ({
                             </Typography>
                         </Menu.Item>
                     )}
-                {/* Uncomment to show project revisions button */}
```
H-Trym, line 201: Should "Project revisions" not be an option in CasesDropMenu? If so should it be removed from CaseDropMenu as well?

H-Trym (approved):
👍

PR #1324, opened 2024-10-22
feat: Navigate to revision after creation

DanielBohme:


frontend/src/Components/Modal/CreateRevisionModal.tsx
```diff
@@ -84,25 +87,32 @@ const CreateRevisionModal: FunctionComponent<Props> = ({
...
+    if (!apiData || !isOpen) { return null }
+
     const internalProjectPhaseOptions = Object.entries(INTERNAL_PROJECT_PHASE).map(([key, value]) => (
-        <option key={key} value={key}>{value.label}</option>
+        <option key={key} value={key} selected={apiData.internalProjectPhase === Number(key)}>{value.label}</option>
     ))
 
     const classificationOptions = Object.entries(PROJECT_CLASSIFICATION).map(([key, value]) => (
-        <option key={key} value={key}>{value.label}</option>
+        <option key={key} value={key} selected={apiData.classification === Number(key)}>{value.label}</option>
```
H-Trym, line 97: using the selected attribute here makes React mad, should use defaultVaule or value instead
image

H-Trym: same for the option on line 93

frontend/src/Components/Controls/ClassificationChip.tsx
```diff
@@ -40,13 +40,12 @@ const SmallTooltip = styled(Tooltip)`
 `
 
 const Classification = () => {
-    const { currentContext } = useModuleCurrentContext()
-    const externalId = currentContext?.externalId
+    const { projectId } = useProjectContext()
```
H-Trym, line 43: This is not related to ClassificationChip, but the revisionChip: 
image
maybe truncate the chip text on smaller screens?

DanielBohme: Should be set as a bug in the devops board

frontend/src/Utils/RevisionUtils.tsx
```diff
@@ -34,6 +22,22 @@ export const navigateToRevision = (
...
+    const newRevision = await projectService.createRevision(projectId, project)
+    if (newRevision) {
+        setCreatingRevision(false)
+    }
+    navigateToRevision(newRevision.id, setIsRevision, queryClient, projectId, navigate)
+}
+
 export const exitRevisionView = (
     setIsRevision: React.Dispatch<React.SetStateAction<boolean>>,
     queryClient: QueryClient,
```
H-Trym, line 43: In RevisionsDropMenu.tsx the setIsRevision(false), seems unnecessary as "exitRevisionView" also sets isRevision to false. 
image

DanielBohme: good catch. will fix

frontend/src/Components/Modal/CreateRevisionModal.tsx
```diff
@@ -84,25 +87,32 @@ const CreateRevisionModal: FunctionComponent<Props> = ({
...
     const submitRevision = () => {
         const newRevision: Components.Schemas.CreateRevisionDto = {
             name: revisionName,
-            internalProjectPhase: internalProjectPhase || apiData.internalProjectPhase,
-            classification: classification || apiData.classification,
+            internalProjectPhase: internalProjectPhase as Components.Schemas.InternalProjectPhase,
+            classification: classification as Components.Schemas.ProjectClassification,
         }
-        return newRevision
+        createRevision(
```
H-Trym, line 108: Have you considered writing the revisionsUtils file as a custom hook instead? This way we would not have to pass the setIsRevision props to the functions

 it can be confusing with the setCreatingRevision being passed from parent to createRevisionModal and from there used in the createRevision function from revisionUtils. 
 
 Rewriting to a hook should be pretty straight forward, would recommend to try out!

H-Trym (approved):
lgtm 👍

PR #1337, opened 2024-10-30
Refactor study cost

taustad:


frontend/src/Components/Case/Components/CaseEditHistory.tsx
```diff
@@ -98,6 +99,10 @@ const CaseEditHistory: React.FC<CaseEditHistoryProps> = ({ caseId }) => {
         }
     }, [activeEdit])
 
+    if (!showEditHistory) {
+        return (<></>)
+    }
```
JonasABach, line 104: Codacy warning: Fragments should contain more than one child - otherwise, there’s no need for a Fragment at all.

frontend/src/Context/AppContext.tsx
```diff
@@ -37,6 +38,8 @@ interface AppContextType {
 const AppContext = createContext<AppContextType | undefined>(undefined)
 
 const AppContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
+    const [showEditHistory, setShowEditHistory] = useState<boolean>(false)
```
JonasABach, line 41: Codacy: 'setShowEditHistory' is assigned a value but never used.

PR #1341, opened 2024-10-31
Fix revision chip messing with the header

H-Trym:


frontend/src/Components/Case/OverviewCasesTable/CasesAgGridTable.tsx
```diff
@@ -103,7 +103,7 @@ const CasesAgGridTable = ({
     const { data: apiRevisionData } = useQuery({
         queryKey: ["revisionApiData", revisionId],
         queryFn: () => revisionQueryFn(projectId, revisionId),
-        enabled: !!externalId && isRevision,
+        enabled: !!projectId && !!revisionId && !!externalId && isRevision,
```
H-Trym, line 106: This query made a console error because it was trying to run without having a projectId or revisionId, adding these to enabled here makes the query not run unless these values are present

frontend/src/Components/Controls/Controls.tsx
```diff
@@ -43,7 +43,7 @@ const Controls = () => {
     const { data: revisionData } = useQuery({
         queryKey: ["revisionApiData", revisionId],
         queryFn: () => revisionQueryFn(projectId, revisionId),
-        enabled: !!revisionId,
+        enabled: !!revisionId && !!projectId,
     })
```
H-Trym, line 47: Same as the console error above, this query was trying to run without the projectId

PR #1345, opened 2024-11-06
fix: Set correct project phase and description after DG0 and disable submit if no revision name

DanielBohme:


kvam (commented):
LGTM, men sett gjerne enda en reviewer på PRen, @DanielBohme.

taustad (approved):

PR #1352, opened 2024-11-07
Bump nuget

kvam:
Bump everything except Microsoft graph (which has api changes in the newest version, will handle this separately)


JonasABach (approved):
lgtm

PR #1372, opened 2024-11-22
Revision does not exit if you change project

JonasABach:


H-Trym (changes requested):
Found a bug related to slow internet speeds while going through this with Jonas

PR #1382, opened 2024-11-28
Revision tabs bug

H-Trym:


H-Trym:
As we have moved away from having revision and project data be separate entities, this PR is no longer needed.

PR #1420, opened 2024-12-19
fix: double click revisions menu to navigate

arian-garshi:


frontend/src/Components/Controls/RevisionsDropMenu.tsx
```diff
@@ -13,25 +13,38 @@ import { useDataFetch } from "@/Hooks/useDataFetch"
...
 type RevisionsDropMenuProps = {
     isMenuOpen: boolean
-    setIsMenuOpen: (isMenuOpen: boolean) => void
+    setIsMenuOpen: (isOpen: boolean) => void
     menuAnchorEl: HTMLElement | null
     isCaseMenu: boolean
 }
 
 interface Revision {
-    revisionId: string
```
H-Trym, line 22: In this app there's a lot of different Ids, Frank has gone through the frontend to try to add some context to the Ids, so that instead of id it should be revisionId or projectId

frontend/src/Components/Controls/RevisionsDropMenu.tsx
```diff
@@ -13,25 +13,38 @@ import { useDataFetch } from "@/Hooks/useDataFetch"
 
 type RevisionsDropMenuProps = {
     isMenuOpen: boolean
-    setIsMenuOpen: (isMenuOpen: boolean) => void
```
H-Trym, line 16: IsMenuOpen might be more readable than isOpen

frontend/src/Components/Controls/RevisionsDropMenu.tsx
H-Trym: being inside the interface "Revision" might be visible enough, but we've tried to make variables more readable

H-Trym (approved):
🗡️

kvam (approved):

PR #1533, opened 2025-02-17
Fix setTimeSeriesData after tableYears instead of after apiData and revisionAndProjectData to prevent race condition

JonasTreD:


H-Trym (approved):
looks good!

PR #1547, opened 2025-02-25
Edit mode read access

JonasTreD:
* Included check to isEditDisabled at places where editMode is checked
* Renamed editMode function parameter to editAllowed for some functions. editAllowed should be editMode && !isEditDisabled
* Show view mode when user only has view access
#60673


frontend/src/Components/Tables/ProjectTables/PROSPCaseTable.tsx
```diff
@@ -165,7 +165,7 @@ const PROSPCaseList = ({
                 label=""
                 value={fileId}
                 onChange={(e: ChangeEvent<HTMLSelectElement>) => handleFileChange(e, p)}
-                disabled={isEditDisabled || !editMode}
+                disabled={!canEdit()}
```
H-Trym, line 168: useCanUserEdit also return isEditDisabled, is it more correct semantically to use that instead of !canEdit() here?

H-Trym: Potentially on all disabled props?

frontend/src/Components/Tables/CaseTables/CaseTabTableWithGrouping.tsx
```diff
@@ -206,9 +208,13 @@ const CaseTabTableWithGrouping = ({
     }, [tableYears])
 
     const handleCellClicked = (event: CellClickedEvent) => {
-        if (!event.data || editMode) { return } // Don't open sidesheet in edit mode
+        if (!event.data || canEdit()) {
```
H-Trym, line 211: Might make sense to still use editMode in this particular case, as the sidesheet should not open in editMode and I'm not sure if userActions.canEditProjectData should have anything to say about wether or not the sidesheet should open?

JonasTreD: if you are in edit mode, but do not have access to actually editing the current project, it should act as you are in view mode. That is what i have implemented, so in my opinion it makes sense to be able to access sidesheet?

You are not able to change the edit mode when in a project you only have view access to

frontend/src/Components/Tables/CaseTables/CaseTabTable.tsx
```diff
@@ -258,10 +260,10 @@ const CaseTabTable = memo(({
             sortable: true,
             filter: true,
             resizable: true,
-            editable: (params: any) => tableCellisEditable(params, editMode),
+            editable: (params: any) => tableCellisEditable(params, canEdit()),
             onCellValueChanged: handleCellValueChange,
             suppressHeaderMenuButton: true,
-            enableCellChangeFlash: editMode,
+            enableCellChangeFlash: canEdit(),
```
H-Trym, line 266: Might be more correct to let the cellChangeFlash be tied only to editMode, as userAccess should effect wether or not the cell should flash on change?

H-Trym: We could potentially expose editMode directly from the new useCanUserEditHook, so we don't have to have two imports for editMode and canEdit (in the cases where the component doesn't already use useAppStore hook for other purposes)

frontend/src/Components/Tables/CaseTables/CaseTabTable.tsx
```diff
@@ -185,24 +187,24 @@ const CaseTabTable = memo(({
...
                     textAlign: "right",
                 },
-                cellClass: (params: any) => (editMode && tableCellisEditable(params, editMode) ? "editableCell" : undefined),
+                cellClass: (params: any) => (tableCellisEditable(params, canEdit()) ? "editableCell" : undefined),
                 valueParser: (params: any) => numberValueParser(setSnackBarMessage, params),
             })
         }
         return columnPinned.concat([...yearDefs])
-    }, [tableYears, editMode, gridRowData, tableName, totalRowName])
+    }, [tableYears, editMode, gridRowData, tableName, totalRowName, isEditDisabled])
```
H-Trym, line 207: Should editMode be removed from dependency array here? (unsure if canEdit should be added as well, it might trigger unnecessary re-renders, but as it's used inside the useCallback it is recommended to have in the dep array I believe

frontend/src/Components/Tables/CaseTables/CaseTabTable.tsx
```diff
@@ -114,9 +116,9 @@ const CaseTabTable = memo(({
     const gridRowData = useMemo(
         () => {
             if (!presentedTableData?.length) { return [] }
-            return profilesToRowData(presentedTableData, dg4Year, tableName, editMode)
+            return profilesToRowData(presentedTableData, dg4Year, tableName, canEdit())
         },
-        [presentedTableData, editMode, dg4Year, tableName],
+        [presentedTableData, editMode, dg4Year, tableName, isEditDisabled],
```
H-Trym, line 121: it looks like editMode can be removed from dependency array here, also canEdit should maybe be added

frontend/src/Components/Tables/CaseTables/CaseTabTable.tsx
JonasTreD: Yeah, may be possible to replace editMode and isEditDisabled, i am just unsure how functions work with dependency array, would this trigger when the result of canEdit is changed? 
if so, i will remove editMode and isEditDisabled and add canEdit instead

frontend/src/Components/Tables/CaseTables/CaseTabTable.tsx
JonasTreD: aah, yeah that sounds like a good idea, will do that

frontend/src/Components/Case/Tabs/CaseDescriptionTab.tsx
```diff
@@ -69,9 +69,9 @@ const CaseDescriptionTab = () => {
             <Gallery />
             <Grid container size={12} justifyContent="flex-start">
                 <Grid container size={{ xs: 12, md: 10, lg: 8 }} spacing={2}>
-                    <Grid size={12} sx={{ marginBottom: editMode ? "32px" : 0 }}>
+                    <Grid size={12} sx={{ marginBottom: canEdit() ? "32px" : 0 }}>
```
H-Trym, line 72: Might be more correct to let the margin be tied only to editMode, as userAccess maybe shouldn't effect this?

frontend/src/Components/Case/Tabs/CaseDrillingSchedule/CaseDrillingScheduleTable.tsx
```diff
@@ -152,11 +154,11 @@ const CaseDrillingScheduleTabTable = ({
     }), [assetWells])
 
     useEffect(() => {
-        const newRowData = wellsToRowData(assetWells, wells, dg4Year, editMode, resourceId, isExplorationTable)
+        const newRowData = wellsToRowData(assetWells, wells, dg4Year, canEdit(), resourceId, isExplorationTable)
         setRowData(newRowData)
         const newColDefs = generateTableYearColDefs()
         setColumnDefs(newColDefs)
-    }, [assetWells, tableYears, wells, editMode])
+    }, [assetWells, tableYears, wells, editMode, isEditDisabled])
```
H-Trym, line 161: remove editMode from dep array? possibly add canEdit

frontend/src/Components/Controls/ProjectControls.tsx
```diff
@@ -131,13 +131,13 @@ const ProjectControls = ({ projectLastUpdated, handleEdit }: props) => {
...
                         >
-                            {editMode && (
+                            {canEdit() && (
                                 <>
                                     <Icon data={visibility} />
                                     {!isSmallScreen && <span>View</span>}
                                 </>
                             )}
-                            {!editMode && (
+                            {(!canEdit()) && (
```
H-Trym, line 140: Shows the edit button if the app is not in edit mode and if the user does not have access to edit. I'm unsure if we currently disable this button if user has no access, we should look into it maybe?

PR #1554, opened 2025-02-27
fix: use correct queryKey for projectApiData

JonasTreD:
useDataFetch uses externalID for projectApiData (fusionProjectID)

* All invalidateQueries for projectApiData now use fusionProjectId
* useRevision no longer invalidates projectApiData

#60675


JonasTreD:
Changed the getRevisionscontroller as it does not need projectId
Improved access management for revisions in middleware
#60830

PR #1573, opened 2025-03-11
fix: loading on image description, and fixing save on case description

H-Trym:


frontend/src/Components/Gallery/Gallery.tsx
```diff
@@ -41,80 +44,37 @@ const Gallery = () => {
...
-                    setExeededLimit(false)
-                } else {
-                    console.error("Image not found for the provided URL:", imageId)
-                }
-            }
-        } catch (error) {
-            console.error("Error deleting image:", error)
+        if (projectId) {
+            deleteImage(projectId, imageId, caseId)
+            // Optimistically update the UI
```
JonasTreD, line 75: any reason why we optimistically updat the ui, instead of relying on the useEffect on line 92?
Could maybe also make this use effect set the exceeded limit instead?

i.e.
    useEffect(() => {
        if (images) {
            setGallery(images)
            setExeededLimit(images.length > 4)
        }
    }, [images])

PR #1578, opened 2025-03-13
Fix: cleanup Gallery and ImageUpload

JonasTreD:
* Bugfix description only allowed for one letter updates, re-add handleDescriptionChange to update gallery when typing image description
* Remove setGallery and SetExeededLimit from ImageUpload, this is handled in gallery using useEffect with image dependency
* ImageUpload now use the useEditGallery (same as as deletes and description updates)
* Removed projectName from upload call to backend, as it is not in use


H-Trym (approved):
lgtm 🗡️

